PR #5, opened 2020-11-26
[DO NOT MERGE] Sample structure for an infra feature

lhl2617:
**_IMPORTANT FOR INFRA TEAM_**

TLDR: Understand all of the code changes in this diff and you're good to go.

This is a rough example of how you would make a code change for a part of feature you own. 
This example is for foraging, and there are two sets of rules (split rule and forage return rule) that can be "set".

This also shows you can "mock" stuff (see `mockClient`) to test interface implementations. It also has a general pattern for code testing, use of enums (called `iota` in go) etc.

Please see the logs in CI for more info.

Test plan:
```
$ go run .
2020/11/26 01:59:08 [SERVER]: Starting forage round!
2020/11/26 01:59:08 [SERVER]: Team 1 invests 42 resources into foraging
2020/11/26 01:59:08 [SERVER]: Team 2 invests 43 resources into foraging
2020/11/26 01:59:08 [SERVER]: Team 3 invests 44 resources into foraging
2020/11/26 01:59:08 [SERVER]: Team 4 invests 45 resources into foraging
2020/11/26 01:59:08 [SERVER]: Team 5 invests 46 resources into foraging
2020/11/26 01:59:08 [SERVER]: Team 6 invests 47 resources into foraging
2020/11/26 01:59:08 [SERVER]: Total forage investments: 267
2020/11/26 01:59:08 [SERVER]: Total payoff from forage: 71289 (using rule: 0)
2020/11/26 01:59:08 [SERVER]: Split of payoff: map[1:11881 2:11881 3:11881 4:11881 5:11881 6:11881]
```

Unit tests pass


internal/clients/team1/client.go
```diff
@@ -32,3 +32,7 @@ func (c *client) GetID() int {
 func (c *client) Logf(format string, a ...interface{}) {
 	log.Printf("[%v]: %v", c.id, fmt.Sprintf(format, a...))
 }
+
+func (c *client) GetForageInvestment(gs common.GameState) int {
```
lhl2617: this unfortunately is something we have to deal with in Golang.

In OOP languages like C++/C# we could have a base class implementing an interface and other classes inheriting from the base class. The base class could by default throw something like a `NotImplementedError`.

When we add stuff into the Client interface, we'd need to manually add in some code to replicate that behaviour; here, I put random numbers, but normally we'd return an error/default value that makes sense (e.g. 0 in this case)

internal/common/forage/payoff.go
```diff
@@ -0,0 +1,40 @@
+package forage
+
+type payoffRuleFuncType = func(input int) (output int)
+
+// DefaultPayoffRuleKey denotes the default payoff for an instance of foraging
+const DefaultPayoffRuleKey = PayoffSquare
```
lhl2617: note the documentation style -- gopls will force you to do this :) This is to make sure documentation is consistent and nice docs can be autogenerated

internal/server/server_test.go
```diff
@@ -0,0 +1,67 @@
...
+// a factory function that allows us to inject custom stuff to facilitate testing
+func somasServerTester(clients []common.Client, gameState common.GameState) Server {
+	return &SOMASServer{
+		clients:   clients,
+		gameState: gameState,
+	}
+}
+
+// extend as required
+type mockClient struct {
```
lhl2617: this is somewhat unsafe, if some test code calls a function we didn't "override" here (we "overrode" `Echo` and `GetID` below), then a segfault will happen. This is test code, so if we safely use it, it should be fine. It is however discouraged, but implementing a complete client is kinda cumbersome--use at your own risk

lhl2617: (`common.Client` is an interface -- I'm forcing mockClient to be an instance of common.Client here)

internal/common/foragepayoff.go
```diff
@@ -0,0 +1,40 @@
+package common
```
lhl2617, line 1: notice the filename here, golang does not allow circular imports, and since we need common stuff here, we can just organise code by filename this way... (some people prefer `forage_payoff.go`, but usually we reserve `_` for build targets or tests.

lhl2617:
closing but can still be referenced

PR #14, opened 2020-11-30
[INFRA/RULES] Added basic implementation of matrix based rules

Neelesh99:
# Summary

Added a basic implementation of Matrix based rule representation. 

## Additional Information

Following from diagram linked here: [https://media.discordapp.net/attachments/778358957902266439/782676119043702814/SOMAS_Rules_2.jpg?width=1038&height=676](url)

"/internal/common/rules/RuleMatrix.go" contains implementation of matrix rules, with global stores of rules and variables. This will likely change to local caches of each within agents.
"/internal/common/rules/ExampleMatrixRegistration.go" contains an example of variable and rule definition, following the example laid out in the above diagram.
"/internal/clients/team3/client.go" contains example fo basic rule evaluation implementation, evaluates the rule defined in diagram above. I've commented out the addition of the evaluation result to the Logging stream (line 22), uncomment this to see the result in the logging stream.

## Test Plan

- `go test ./...`
```
?       github.com/SOMAS2020/SOMAS2020  [no test files]
?       github.com/SOMAS2020/SOMAS2020/internal/clients/team1   [no test files]
?       github.com/SOMAS2020/SOMAS2020/internal/clients/team2   [no test files]
?       github.com/SOMAS2020/SOMAS2020/internal/clients/team3   [no test files]
?       github.com/SOMAS2020/SOMAS2020/internal/clients/team4   [no test files]
?       github.com/SOMAS2020/SOMAS2020/internal/clients/team5   [no test files]
?       github.com/SOMAS2020/SOMAS2020/internal/clients/team6   [no test files]
ok      github.com/SOMAS2020/SOMAS2020/internal/common  0.329s
ok      github.com/SOMAS2020/SOMAS2020/internal/common/rules    0.118s
ok      github.com/SOMAS2020/SOMAS2020/internal/server  0.225s
?       github.com/SOMAS2020/SOMAS2020/pkg/testutils    [no test files]

```

- `go run .`
```
2020/11/30 17:56:02 [0]: Echo: 'Hello World'
2020/11/30 17:56:02 [SERVER]: Received echo `Hello World` from client 0
2020/11/30 17:56:02 [1]: Echo: 'Hello World'
2020/11/30 17:56:02 [SERVER]: Received echo `Hello World` from client 1
2020/11/30 17:56:02 [2]: Echo: 'Hello World'
2020/11/30 17:56:02 [SERVER]: Received echo `Hello World` from client 2
2020/11/30 17:56:02 [3]: Echo: 'Hello World'
2020/11/30 17:56:02 [SERVER]: Received echo `Hello World` from client 3
2020/11/30 17:56:02 [4]: Echo: 'Hello World'
2020/11/30 17:56:02 [SERVER]: Received echo `Hello World` from client 4
2020/11/30 17:56:02 [5]: Echo: 'Hello World'
2020/11/30 17:56:02 [SERVER]: Received echo `Hello World` from client 5

```


internal/common/rules/ExampleMatrixRegistration.go
```diff
@@ -0,0 +1,30 @@
+package rules
```
lhl2617: https://stackoverflow.com/questions/25161774/what-are-conventions-for-filenames-in-go/25162021#25162021

nit for file naming convention :)

internal/common/rules/ExampleMatrixRegistration.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"gonum.org/v1/gonum/mat"
+)
+
+func RegisterCoolRool() {
```
lhl2617: let's follow godoc conventions?
https://blog.golang.org/godoc

internal/common/rules/RuleMatrix.go
```diff
@@ -0,0 +1,48 @@
...
+	RequiredVariables []string
+	ApplicableMatrix  mat.Dense
+	AuxiliaryVector   mat.VecDense
+}
+
+var AvailableRules = map[string]RuleMatrix{}
+
+func registerNewRule(ruleName string, requiredVariables []string, applicableMatrix mat.Dense, auxiliaryVector mat.VecDense) *RuleMatrix {
+	if _, ok := AvailableRules[ruleName]; ok {
+		//Some form of anger at this point
```
lhl2617: you can use `panic(...)`

internal/common/rules/RuleMatrix.go
```diff
@@ -0,0 +1,48 @@
...
+type RuleMatrix struct {
+	ruleName          string
+	RequiredVariables []string
+	ApplicableMatrix  mat.Dense
+	AuxiliaryVector   mat.VecDense
+}
+
+var AvailableRules = map[string]RuleMatrix{}
+
+func registerNewRule(ruleName string, requiredVariables []string, applicableMatrix mat.Dense, auxiliaryVector mat.VecDense) *RuleMatrix {
```
lhl2617: why return the matrix? is this used?

internal/common/rules/RuleMatrix.go
Neelesh99: Sure, does panic like end the program?, cos I was thinking of a less severe penalty for trying to re-register a rule.

internal/common/rules/ExampleMatrixRegistration.go
```diff
@@ -0,0 +1,30 @@
...
+
+	//A very contrived rule//
+
+	registerNewVariable("number_of_islands_contributing_to_common_pool", 5)
+	registerNewVariable("number_of_failed_forages", 0.5)
+	registerNewVariable("number_of_broken_agreements", 1)
+	registerNewVariable("max_severity_of_sanctions", 2)
+
+	name := string("Kinda Complicated Rule")
```
lhl2617: i think you can skip the `string(...)`

internal/common/rules/RuleMatrix.go
Neelesh99: It depends on how our Global/Local rule caching architecture works. I returned it for completeness but can take it out if we are going for global cache.

internal/common/rules/ExampleMatrixRegistration.go
Neelesh99: Lol GoLand warned me and I did not listen, removing that now

internal/common/rules/ExampleMatrixRegistration.go
```diff
@@ -0,0 +1,30 @@
...
+	reqVar := make([]string, 4)
```
lhl2617:
```golang
reqVar := []string{
     "number_of_islands_contributing_to_common_pool",
     ....
}
```
should work.

These variables should also be `enum`s IMO? (in golang they're `iota`s that instead return `int`s, but we could just have const string literals

lhl2617: They should be exported as well for clients to read it. we can have a file with just these constants

Neelesh99: Yeah tbh enum would be a lot cleaner, I wasn't thinking that far ahead tbh, and also with the self-generated rules I wasn't sure how that works.

lhl2617: ah--i just saw these being used in `registerNewVariable`. I think having them as exported constnats in a separate file (rules are bound to have shared variables at some point I think?) would be better

internal/common/rules/RuleMatrix.go
```diff
@@ -0,0 +1,48 @@
...
+		//Some form of anger at this point
+	}
+
+	VariableMap[variableName] = value
+}
+
+func modifyVariable(variableName string, newValue float64) {
+	if _, ok := VariableMap[variableName]; ok {
+		VariableMap[variableName] = []float64{newValue}
+	}
```
lhl2617: else return an `error` I suppose (if this is exposed to clients)

internal/common/rules/RuleMatrix.go
lhl2617: ...or return an error, but since this is something we want to do instantly at startup in infra, panic is encouraged

lhl2617: yep, panic ends it, or you can return an `error` and let the caller handle it.

internal/clients/team3/client.go
```diff
@@ -1,34 +1,93 @@
...
+		case 0:
+			resultVect = append(resultVect, c.AtVec(i) == 0)
+		case 1:
+			resultVect = append(resultVect, c.AtVec(i) > 0)
+		case 2:
+			resultVect = append(resultVect, c.AtVec(i) >= 0)
+		case 3:
+			resultVect = append(resultVect, c.AtVec(i) != 0)
+		}
+	}
```
lhl2617: should this logic be in client or common code?

internal/clients/team3/client.go
```diff
@@ -1,34 +1,93 @@
...
+
+	for _, v := range variables {
+		variableVect = append(variableVect, rules.VariableMap[v]...)
+	}
+
+	variableVect = append(variableVect, 1)
+
+	variableFormalVect := mat.NewVecDense(len(variables)+1, variableVect)
+
+	rows, _ := rm.ApplicableMatrix.Dims()
```
lhl2617, line 38: ignoring error?? :(

internal/common/rules/RuleMatrix.go
Neelesh99: Sure, done

internal/clients/team3/client.go
lhl2617:
```golang
resultVect := []bool {
    c.AtVec(0) == 0,
    c.AtVec(1) > 0,
    ...
}
```

Neelesh99: So this is where we have to make a decision whether to give all clients a basic version of the Matrix evaluator in the common client file, or whether to tell them to implement it themselves using a provided spec. I've put it in a specific client for now because it's less disruptive to change this code while we are undecided on that.

lhl2617:
```golang
	for i := 0; i < rows; i++ {
                res := false
		switch interpret := aux.AtVec(i); interpret {
		case 0:
			res = c.AtVec(i) == 0
		...
                default:
                       // return an error? once this is factored out in a function
                 resultVect = append(resultVect, res)
	}
```

internal/clients/team3/client.go
```diff
@@ -1,34 +1,93 @@
...
+	for _, v := range resultVect {
+		finalBool = finalBool && v
+	}
+
+	return finalBool
+}
+
+func (c *client) Echo(s string) string {
+
+	rules.RegisterCoolRool()
```
lhl2617: this should be done in `common` I think? You can put these in an [init](https://medium.com/golangspec/init-functions-in-go-eac191b3860a) function

lhl2617:
Rly cool -- I'm on board. Can we segregate logic into smaller functions and have unit tests? (like the interpretation of the results I would love to see tested and segregated). Other reps pls have a look?

Neelesh99 (commented):
files renamed

internal/common/rules/ExampleMatrixRegistration.go
Neelesh99: as discussed we'll stick to strings for now

internal/common/rules/globalvariablecache.go
```diff
@@ -0,0 +1,30 @@
+package rules
+
+import "errors"
+
+type VariableValuePair struct {
+	VariableName string
+	Multivalued  bool
+	SingleValue  float64
+	MultiValue   []float64
```
lhl2617: wouldn't SingleValue be implied from `Values []float64` having one value?

internal/common/rules/globalvariablecache.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+var VariableMap = map[string]VariableValuePair{}
+
+func RegisterNewVariable(pair VariableValuePair) error {
+	if _, ok := VariableMap[pair.VariableName]; ok {
+		return errors.New("attempted to re-register a variable that had already been registered")
```
lhl2617, line 15: noice

internal/common/rules/globalvariableregistration.go
```diff
@@ -0,0 +1,37 @@
+package rules
+
+func init() {
+	for _, v := range StaticVariables {
+		e := RegisterNewVariable(v)
+		if e != nil {
+			panic("variable registration gone wrong, variable " + v.VariableName + " has been registered multiple times")
```
lhl2617: yeeet!

consider fstrings:

`fmt.Sprintf("Variable registration gone strong: '%v' already registered", v.VariableName")`

internal/clients/team3/client.go
Neelesh99: Yup done

internal/common/rules/globalvariablecache.go
Neelesh99: Yeah idk why I even tried this tbh XD, fixed now

internal/common/rules/globalvariableregistration.go
Neelesh99: Yup fixed

internal/common/rules/ExampleMatrixRegistration.go
Neelesh99: Added in latest commit

internal/clients/team3/client.go
```diff
@@ -1,34 +1,44 @@
...
+func (c *client) Echo(s string) string {
+
+	//c.DemoEvaluation()
+	c.Logf("Echo: '%v'", s)
+
+	return s
+}
+
+func (c *client) DemoEvaluation() {
+	getEval, err := common.BasicRuleEvaluator("Kinda Complicated Rule")
```
lhl2617: nit: `getEval` sounds like a name for a getter function instead.

internal/common/client.go
```diff
@@ -1,44 +1,117 @@
...
+		}
+
+		var finalBool = true
+
+		for _, v := range resultVect {
+			finalBool = finalBool && v
+		}
+
+		return finalBool, nil
+	} else {
```
lhl2617, line 114: nit: usually we have shorter cases on top (easier to read), but this is perfectly fine, no worries.

internal/common/client.go
```diff
@@ -1,44 +1,117 @@
...
+	}
+	RegisteredClients[id] = c
+}
+
+// BasicRuleEvaluator implements a basic version of the Matrix rule evaluator, provides single boolean output (and error if present)
+func BasicRuleEvaluator(ruleName string) (bool, error) {
+	if rm, ok := rules.AvailableRules[ruleName]; ok {
+		variables := rm.RequiredVariables
+
+		var variableVect []float64
```
lhl2617: move this above L54 so we can immediately see the for loop is iterating through `rm.RequiredVariables`?

internal/common/client.go
```diff
@@ -1,44 +1,117 @@
...
+				variableVect = append(variableVect, val.Values...)
+			} else {
+				return false, errors.New(fmt.Sprintf("Variable: '%v' not found in global variable cache", v))
+			}
+		}
+
+		variableVect = append(variableVect, 1)
+
+		//Checking dimensions line up
+		rows, cols := rm.ApplicableMatrix.Dims()
```
lhl2617: nit: this is fine, but I think `rows` and `cols` might mislead people to think that it refers to the actual slices/matrices of `rows` and `cols`.

How about `nRows` or `numRows`?

internal/common/client.go
```diff
@@ -1,44 +1,117 @@
...
+
+		if cols != len(variableVect) {
+			return false, errors.New(
+				fmt.Sprintf("dimension mismatch in evaluating rule: '%v' rule matrix has '%v' columns, while we sourced '%v' variables",
+					ruleName,
+					strconv.Itoa(cols),
```
lhl2617: `strconv` not needed :) `%v` is smart.

internal/common/client.go
```diff
@@ -1,44 +1,117 @@
...
+		actual := make([]float64, rows)
+
+		c := mat.NewVecDense(rows, actual)
+
+		c.MulVec(&rm.ApplicableMatrix, variableFormalVect)
+		aux := rm.AuxiliaryVector
+
+		var resultVect []bool
+
+		for i := 0; i < rows; i++ {
```
lhl2617: probably a good candidate to be moved out in another function? quite a long of lines in this `if` body :)

internal/common/client.go
```diff
@@ -1,44 +1,117 @@
...
+			res := false
+			switch interpret := aux.AtVec(i); interpret {
+			case 0:
+				res = c.AtVec(i) == 0
+			case 1:
+				res = c.AtVec(i) > 0
+			case 2:
+				res = c.AtVec(i) >= 0
+			case 3:
+				res = c.AtVec(i) != 0
```
lhl2617, line 100: let's catch default errors as well -- would be hard to debug when you have a "falsey-false" in your `resultVect` due to a problem with `interpret` not being in `[0, 3]`

lhl2617: ...and causing `finalBool` to be falsely false

internal/common/client_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"github.com/SOMAS2020/SOMAS2020/internal/common/rules"
+	"gonum.org/v1/gonum/mat"
+	"testing"
+)
+
+// TestBasicRuleEvaluatorPositive Checks whether rule we expect to evaluate as true actually evaluates as such
+func TestBasicRuleEvaluatorPositive(t *testing.T) {
+	result, err := BasicRuleEvaluator("Kinda Complicated Rule")
```
lhl2617, line 11: Hmm... this test relies on the registration of `"Kinda Complicated Rule"`. which we might change. It's nice to have self-contained unit tests that don't rely on other (possibly side-effecty) code.

w.r.t. this file we could have the logic for `registerTestRule` contained in the `TestBasicRuleEvaluatorNegative` function as well. 

Thanks for the unit tests though! Tests are always hard to write!

lhl2617: Also this might only be important in the final PR, but having them self-contained also means you can have one const. variable referring to supposed constants like the rule name :)

internal/common/rules/globalruleregistration_test.go
```diff
@@ -0,0 +1,16 @@
+package rules
+
+import "testing"
+
+// TestGlobalRuleRegistration checks whether required rules are registered
+func TestGlobalRuleRegistration(t *testing.T) {
+	rulesToFind := []string{
+		"Kinda Complicated Rule",
```
lhl2617, line 8: fine for now, but would be nice if we have the file of constants for the names of rules as well, then we can have a nice self-contained unit test for `AvailableRules` testing for all intended rules.

lhl2617:
also cc other reps: I know I'm annoying at code reviews! But if we have good quality (hopefully) it'll make everyone's lives easier :)

go.sum
```diff
@@ -0,0 +1,50 @@
...
+golang.org/x/tools v0.0.0-20201125231158-b5590deeca9b/go.mod h1:emZCQorbCU4vsT4fOWvOPXz4eW1wZW4PmDk9uLelYpA=
+golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
+golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
+golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
+gonum.org/v1/gonum v0.0.0-20180816165407-929014505bf4/go.mod h1:Y+Yx5eoAFn32cQvJDxZx5Dpnq+c3wtXuadVZAcxbbBo=
+gonum.org/v1/gonum v0.8.1 h1:wGtP3yGpc5mCLOLeTeBdjeui9oZSz5De0eOjMLC/QuQ=
+gonum.org/v1/gonum v0.8.1/go.mod h1:oe/vMfY3deqTw+1EZJhuvEW2iwGF1bW9wwu7XCu0+v0=
+gonum.org/v1/netlib v0.0.0-20190313105609-8cb42192e0e0/go.mod h1:wa6Ws7BG/ESfp6dHfk7C6KdzKA7wR7u/rKwOGE66zvw=
+gonum.org/v1/plot v0.0.0-20190515093506-e2840ee46a6b/go.mod h1:Wt8AAjI+ypCyYX3nZBvf6cAIx93T+c/OS2HFAYskSZc=
+rsc.io/pdf v0.1.1/go.mod h1:n8OzWcQ6Sp37PL01nO98y4iUCRdTGarVfzxY20ICaU4=
```
QFSW, line 50: What's all this? Auto generated?

internal/common/client_test.go
```diff
@@ -0,0 +1,44 @@
...
+	if !result {
+		t.Errorf("Rule evaluation came as false, when it waqs expected to be true, potential error with value '%v'", err)
```
QFSW: typo in was

internal/common/client_test.go
```diff
@@ -0,0 +1,44 @@
...
+		t.Errorf("Rule evaluation came as true, when it was expected to be false, potential error with value '%v'", err)
+	}
+}
+
+func registerTestRule() {
+
+	//A very contrived rule//
+	name := "Kinda Test Rule"
+	reqVar := []string{
+		"number_of_islands_contributing_to_common_pool",
```
QFSW, line 31: Perhaps we should change the "magic strings" defining the rule components with an enum?

internal/common/rules/globalrulescache.go
```diff
@@ -0,0 +1,16 @@
+package rules
+
+import "gonum.org/v1/gonum/mat"
+
+var AvailableRules = map[string]RuleMatrix{}
+
+// RegisterNewRule Creates and registers new rule based on inputs
+func RegisterNewRule(ruleName string, requiredVariables []string, applicableMatrix mat.Dense, auxiliaryVector mat.VecDense) *RuleMatrix {
+	if _, ok := AvailableRules[ruleName]; ok {
+		panic("attempt to re-register rule with name " + ruleName)
```
QFSW: We should use an error code rather than a panic because a bad client shouldn't be able to crash the simulation

internal/common/client_test.go
lhl2617: as per https://github.com/SOMAS2020/SOMAS2020/pull/14/files#r533407903 :)

In convo with @Neelesh99  i think eventually these will be constants in a file

internal/common/rules/globalrulescache.go
lhl2617: agreed - if clients can call this we should error instead of panic

internal/clients/team3/client.go
Neelesh99: Yup, fair enough, fixed

internal/common/client.go
Neelesh99: Fair enough, thanks

internal/common/client.go
Neelesh99: Yup lol, thanks

internal/common/client.go
Neelesh99: Thanks for the tip

internal/common/client.go
Neelesh99: Will work more on this in a later PR

internal/common/client.go
Neelesh99: Fair enough

internal/common/rules/globalruleregistration_test.go
Neelesh99: Yup will be doing an overhaul of unit tests in later PR

go.sum
Neelesh99: Lol yeah, not sure what it is, LH said it was ok to leave in

internal/common/rules/globalrulescache.go
Neelesh99: Yup agreed and changed

internal/clients/team3/client.go
```diff
@@ -1,34 +1,44 @@
...
+	return s
+}
+
+func (c *client) DemoEvaluation() {
+	evalResult, err := common.BasicRuleEvaluator("Kinda Complicated Rule")
+	if err != nil {
+		panic(err.Error())
+	}
+	c.Logf("Rule Eval: %t", evalResult)
+}
```
lhl2617, line 34: OK, we'll eventually have to change the interface of `Client` to have a rule evaluator I presume?

internal/common/client_test.go
lhl2617: @Neelesh99  quick fix? 👍

Neelesh99: Ty

internal/clients/team3/client.go
Neelesh99: Yeah in a bit, once we have a bit more of a design decision on the local vs global caches for variables and stuff

lhl2617 (approved):
image

OK, let's iterate on this as we go along! Good job @Neelesh99  and team! 
@QFSW quick if you wanna have a final look... :)

PR #28, opened 2020-12-03
[INFRA/CORE] Fixed file exclusion in protected action

QFSW:
# Summary

Fixed file exclusion in protected action


lhl2617 (approved):
image

PR #53, opened 2020-12-16
[INFRA/CORE] Run golangci-lint more verbosely

lhl2617:
# Summary

Add `-min-confidence=0` argument to catch more lint errors


QFSW:

PR #79, opened 2020-12-20
Local GameState Variables Added to BaseClient

preetl:
# Summary

Added variables to store ClientInfo, Turn and Season from gameState on the BaseClient.

## Additional Information

:)

## Test Plan

N/A

### DIFF
--- a/internal/common/baseclient/baseclient.go
+++ b/internal/common/baseclient/baseclient.go
@@ -35,6 +35,7 @@ func NewClient(id shared.ClientID) Client {
 // even when new features are added.
 type BaseClient struct {
 	id shared.ClientID
+	clientInfo ClientInfo
 }
 
 // Echo prints a message to show that the client exists
@@ -69,5 +70,6 @@ func (c *BaseClient) StartOfTurnUpdate(gameState gamestate.ClientGameState) {
 // The gameState is served by the server.
 // OPTIONAL. Base should be able to handle it but feel free to implement your own.
 func (c *BaseClient) GameStateUpdate(gameState gamestate.ClientGameState) {
+	c.clientInfo = gameState.ClientInfo.Copy()
 	c.Logf("Received game state update: %#v", gameState)
 }
--- a/internal/common/baseclient/baseclient.go
+++ b/internal/common/baseclient/baseclient.go
@@ -35,7 +35,9 @@ func NewClient(id shared.ClientID) Client {
 // even when new features are added.
 type BaseClient struct {
 	id shared.ClientID
-	clientInfo ClientInfo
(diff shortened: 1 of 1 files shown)


QFSW:

Neelesh99 (approved):
Looks good to me

internal/common/baseclient/baseclient.go
```diff
@@ -69,5 +72,8 @@ func (c *BaseClient) StartOfTurnUpdate(gameState gamestate.ClientGameState) {
 // The gameState is served by the server.
 // OPTIONAL. Base should be able to handle it but feel free to implement your own.
 func (c *BaseClient) GameStateUpdate(gameState gamestate.ClientGameState) {
+	c.clientInfo = gameState.ClientInfo.Copy()
```
lhl2617, line 75: `.Copy` not needed :) -- it was already passed in as a copy.

Can't the turn and season be acquired from `c.clientInfo.Turn` as well? I don't see the need of the two extra `uint` vars.

Do we need this for `StartOfTurnUpdate` as well?

preetl: whoops didn't notice these comments! yes, we do need it for StartOfTurnUpdate too - will do that now.

preetl: @lhl2617 i don't see that clientInfo has Turn and Season..? It's just Resources, LifeStatus and Critical...

lhl2617: Whoops sorry! yeah it doesnt :) -- how about just defining a `clientGameState` var in the struct then? :)

preetl: ah ok sure - shall i make a copy function in clientgamestate? i don't think i get when we need a copy func and when not ...

lhl2617: Good question! 

Outside of this code, the short answer is "same as Java and Python". Some types are *primitive*, and they are always _passed by copy_ (such as ints, strings, floats). Other types are *reference types* and they are _passed by value of the reference_ (in golang: maps, slices, channels, functions). 

In Python we'd need to `deepcopy` the reference types to obtain a true copy, similar here.

As when the server calls `GameStateUpdate` the ClientGameState is already (deep-)copied, you don't need to call it anywhere in client side.

lhl2617: The risk of not copying is that the reference passed to the client means that the client has access to change the reference types to some bogus value. So all copying should be done _before_ passing it to the client.

preetl: ahhhhh, i see - thank you :) 
i'll make a new pr seeing as this already merged

PR #80, opened 2020-12-20
[INFRA/CORE] Add note on perm denied log file creation

lhl2617:
# Summary

`sudo` _might be_ required to create a new file on macOS because we're using absolute paths.

## Additional Information
For once, Windoze wins.
Removed `sudo` in CI--not needed as permissions are now `0777`

## Test Plan
¯\\_(ツ)_/¯

Tested with @preetl


QFSW:

preetl (approved):
lovely :)

PR #86, opened 2020-12-21
Add some functions about voting

Ning-72:


QFSW:

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
+package vote
+
+import (
+	"math/rand"
+	"time"
+)
+
+func Get_votes_for_rule(ruleID string, islandID int) map[int][]int {
```
lhl2617:
```suggestion
func GetVotesForRule(ruleID string, islandID int) map[int][]int {
```
Golang does not encourage snake case :)

lhl2617: What does this function do? Can you document it please?

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
...
+	votes_layout := make(map[int][]int)
+	rand.Seed(time.Now().UTC().UnixNano())
+	i := 1
+	for {
+		votes_layout[i] = []int{rand.Intn(100), rand.Intn(100)}
+		i++
+		if i > 6 {
+			break
+		}
+	}
```
lhl2617:
```
votesLayout = make(map[int][]int)
for i := 1; i <= 6; i++ {
    votesLayout[i] = []int{rand.Intn(100), rand.Intn(100)}
}
```

internal/common/vote/vote.go
lhl2617: If this function is supposed to reach out to the island, it should return `map[shared.ClientID][]int`

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
+package vote
```
lhl2617, line 1: We use `shared.ClientID` to refer to islands/clients--please change to reflect that? :) Don't use 1-6

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
...
+			results_of_all_islands[j] = []int{1, 0}
+		}
+		if votes_layout[j][0] <= votes_layout[j][1] {
+			results_of_all_islands[j] = []int{0, 1}
+		}
+		j++
+		if j > 6 {
+			break
+		}
+	}
```
lhl2617: use the `for i := 1; i <= 6; i++` syntax? It's clearer

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
...
+		if v[1] == 1 {
+			num_rej++
+		}
+	}
+	if num_acc > num_rej {
+		ans = true
+	}
+	if num_acc <= num_rej {
+		ans = false
+	}
```
lhl2617: this seems like simple majority? `VoteRule` does not imply that.

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
...
+	return num_acc, num_rej, ans
+}
+
+func Get_votes_for_elect() map[int][]int {
```
lhl2617: snake case to pascal case please :)

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
...
+
+	//Get votes from each island for the election.
+	votes_layout_e := make(map[int][]int)
+	votes_layout_e = Get_votes_for_elect()
+
+	//Calculate the preference map.
+	var order [6]int
+	var index [6]int
+	var score [6]int
+	PreferenceMap := make(map[int][]int)
```
lhl2617: lowercase `p` for local variables.

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
...
+	for k, v := range votes_layout_e {
```
lhl2617: can you break this down into smaller pieces so we can understand what's going on? :) also use the `for i := 0; i <= x; i++` syntax :)

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
...
+
+func VoteElectSpeaker() (int, [6]int, map[int][]int) {
+	win, Final_score, PreferenceMap := Vote_Elect()
+	return win, Final_score, PreferenceMap
+}
+
+func VoteElectPresident() (int, [6]int, map[int][]int) {
+	win, Final_score, PreferenceMap := Vote_Elect()
+	return win, Final_score, PreferenceMap
+}
```
lhl2617: these functions seem to all have the same body? maybe make the body into another function and call that from these 3 functions?

lhl2617 (changes requested):
Please use the three-component for loop syntax: https://tour.golang.org/flowcontrol/1 -- it makes code clearer to read :)

Further, use `shared.ClientID` to refer to clients. Also write comments to document your functions so we know how to use it :)

Address also lint errors.

lhl2617:
P/S: Your main branch is also out of date--use the `Update branch`  button to update it.
image

This infra feature should also come from a branch in SOMAS2020, with agent side code coming from your fork.

lhl2617:
@Neelesh99  can you make sure it also fits into your rule implementation? Thanks :D

internal/common/vote/vote.go
Ning-72: It is expected to get voting information from each island, but it seems that how will each island vote under certain circumstances has not been determined by design team, so I directly generate random votes inside the function instead.

lhl2617: Right, thanks :) -- this should be in the comment of the function then, with a `TODO` noting that it should be changed when the island voting interface is added.

Can you also ping the voting design rep to confirm this? I assume we need voting for the islands in the MVP. I'm not sure whose responsibility it is to add it to the client interface.

internal/common/vote/vote.go
QFSW: Don't hardcode the number of islands, we may want to have more than one island per team in the future (e.g. same client code initialized with different values). This should be resolved throughout the PR

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
+package vote
+
+import (
+	"math/rand"
+	"time"
+)
+
+func GetVotesForRule(ruleID string, islandID int) map[int][]int {
+	votes_layout := make(map[int][]int)
```
QFSW: I believe we use camelCase as our naming convention for locals as per go standards, best to keep it consistent. This should be resolved throughout the PR

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
...
+	rand.Seed(time.Now().UTC().UnixNano())
+	i := 1
+	for {
+		votes_layout[i] = []int{rand.Intn(100), rand.Intn(100)}
```
QFSW: Why are we initializing the layout to random integers between 0 and 100? it isn't clear to me from the code what is happening

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
+package vote
+
+import (
+	"math/rand"
+	"time"
+)
+
+func GetVotesForRule(ruleID string, islandID int) map[int][]int {
+	votes_layout := make(map[int][]int)
+	rand.Seed(time.Now().UTC().UnixNano())
```
QFSW: I don't think seeding from within the function body is good practice, the seed should be performed externally during initialisation. This means if we later decide that we want to re simulate with the same seed we are able to do so, and we can also log that seed. This should be resolved throughout the PR

internal/common/vote/vote.go
QFSW: This should also be done throughout the PR

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
...
+
+	//Calculate the final score for six island and ditermine the winner.
+	var Final_score = [6]int{0, 0, 0, 0, 0, 0}
+	for _, v := range PreferenceMap {
+		Final_score[0] = Final_score[0] + v[0]
+		Final_score[1] = Final_score[1] + v[1]
+		Final_score[2] = Final_score[2] + v[2]
+		Final_score[3] = Final_score[3] + v[3]
+		Final_score[4] = Final_score[4] + v[4]
+		Final_score[5] = Final_score[5] + v[5]
```
QFSW: Use a for loop instead of copying and pasting chunks of code

internal/common/vote/vote.go
```diff
@@ -0,0 +1,191 @@
...
+			order[j] = 101
+
+			i++
+			if i > 5 {
+				break
+			}
+		}
+
+		o := 0
+		r := 0
```
QFSW: Outside of some cases (like i for a loop index, n for a count) please avoid using single letter names for variables as they are not descriptive.

QFSW (changes requested):
Please add a description to the PR describing what was changed following the template provided showing that the simulation and tests have been run

Please make sure that your source branch is up to date before merging to main as it helps avoid having to resolve conflicts within the PR itself

I've left some comments about the code throughout the review

Neelesh99:
> @Neelesh99 can you make sure it also fits into your rule implementation? Thanks :D

From what I can see this fits with both Rules and IIGO for mvp

internal/common/vote/vote.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func VoteRule(ruleID string, Numofislands int) (int, int, bool) {
+
+	//Get votes from islands for the rule changes.
+	voteslayout := make(map[int][]int)
+	voteslayout = Getvotesforrule(ruleID, Numofislands)
+
+	//Calculate results of each island.
+	resultsofallislands := make(map[int][]int)
```
QFSW: camelCase means it should be `resultsOfAllIslands` not `resultsofallislands`

internal/common/vote/vote.go
```diff
@@ -0,0 +1,192 @@
...
+	if numacc > numrej {
+		ans = true
+	}
+	if numacc <= numrej {
+		ans = false
+	}
+	return numacc, numrej, ans
+}
+
+func Getvotesforelect(Numofislands int) map[int][]int {
```
QFSW: PascalCase means it should be `GetVotesForElect` not `Getvotesforelect`

QFSW (changes requested):

internal/common/vote/vote.go
```diff
@@ -0,0 +1,192 @@
+package vote
+
+import (
+	"math/rand"
+	"time"
+)
+
+func GetVotesForRule(ruleID string, NumOFIslands int) map[int][]int {
```
QFSW: `NumOfIslands` for PascalCase, not `NumOFIslands `

Ning-72: Sorry...

QFSW: Function arguments should be camelCase, so it should be `numOfIslands`

internal/common/vote/vote.go
```diff
@@ -0,0 +1,192 @@
...
+		ans = true
+	}
+	if numacc <= numrej {
+		ans = false
+	}
+	return numacc, numrej, ans
+}
+
+func GetVotesForElect(NumOFIslands int) map[int][]int {
+	rand.Seed(time.Now().UTC().UnixNano())
```
QFSW: Same as previous comment on seeding

QFSW (changes requested):

internal/common/vote/vote.go
Ning-72: Every time an election take place, the number of ballots will be randomly generated. If we want to run that election once again with the ballot number same as last time, we should record details about past few elections (at least the last one). Do you mean this?

QFSW: When you call `rand.Seed()` you are causing a global state change because you are changing the random number sequence used by all code that uses `rand`. That's okay, but typically that is done by initialization code, and not in the specific function that uses `rand` like done here.

The problem with this approach is that every time this function is called the random number generator is re-seeded. Typically you only want to seed it once

Furthermore, sometimes you want to use a fixed seed. Say for example we wanted to run the exact same simulation twice, in order to do this we would use a fixed value for the seed in the initialization code. The current problem is, when this function is invoked, it will override the random seed making it impossible for us to do that

Long story short, we should call `rand.Seed()` once from main, but not from here (or any of these functions)

internal/common/vote/vote.go
```diff
@@ -0,0 +1,188 @@
+package vote
+
+//will be generated from client
+var voteForRule map[int][]int
+var voteForElect map[int][]int
+
+func GetVotesForRule(ruleID string, numOfIslands int) map[int][]int {
```
Eirikalb: This method should be part of the baseClient.

internal/common/vote/vote.go
```diff
@@ -0,0 +1,188 @@
...
+	if numacc > numrej {
+		ans = true
+	}
+	if numacc <= numrej {
+		ans = false
+	}
+	return numacc, numrej, ans, votesLayoutRule
+}
+
+func GetVotesForElect(numOfIslands int) map[int][]int {
```
Eirikalb: Same for this, the implementation of the vote counting should be on the server side, and then the vote itself should come from the clients

internal/common/vote/vote.go
```diff
@@ -0,0 +1,188 @@
...
+		i++
+		if i > numOfIslands {
+			break
+		}
+	}
+
+	return votesLayoutElect
+}
+
+func VoteElect(numOfIslands int) (int, []int, map[int][]int) {
```
Eirikalb, line 53: The functions should all have comments defining what they actually do. Is this the function that allows the islands to cast a vote?

Eirikalb (changes requested):
AFAICT this does not interact with the rest of the infrastructure yet. Some of these functions should be part of server, and will need to be fitted to the current code. Especially IIGO will be relevant here. Lets have a quick meeting and I can show you how to implement your changes to add voting for the client side, and maybe you could meet with team 3/4 later to figure out where the voting part needs to go on the server side?

internal/common/vote/vote.go
Ning-72: No, this function is just to collect votes from every island and write them in a map. IIGO will send a voting request to islands to my understanding. I think not interacting with other parts of infrastructure is mainly because the only input needed for voting is every island's votes, which is not given by design team at this stage. Once the voting rules are done by design, I guess I could implement it then? Also from my understanding, connection between IIGO and voting means that functions in voting could be called in IIGO, but voting doesn't need any information from IIGO right?

Ning-72: Sorry func VoteElect is to calculate the voting result(who is the winner), func GetVotesForElect is to collect votes from every island and write them in a map, same as VoteRule and GetVotesForRule, I'll add explanation in the code.

Neelesh99:
Little bit of work needed. Will try to integrate this voting with IIGO functions and work with Team6 for refactoring.

PR #89, opened 2020-12-21
IIGO Implementation for MVP

preetl:
# Summary

- Implements role objects: BasePresident, BaseSpeaker and BaseJudge, to meet MVP requirement for compliant islands
- President, Speaker and Judge interfaces to allow overriding by islands Post-MVP
- Extended BaseClient and Client Interface to include 'communication' functions required for IIGO

N.B. IIGO session is conducted sequentially through orchestration.go to allow communication between role objects

Sorry for the mahoosive PR!

## Additional Information

Auxiliary information goes here.

## Test Plan

Unit tests mostly but more testing to come!

PR #90, opened 2020-12-21
[INFRA/CORE] Added .gitattributes file

QFSW:
# Summary

Added a `.gitattributes` file. This is because different systems use different line endings which can sometimes cause large whitespace commits or annoying to resolve merge issues. See pr #89 for what may potentially be a symptom of this.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+* text=auto
+
+*.go text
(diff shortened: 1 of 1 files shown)


QFSW:

preetl (approved):
beautiful!

PR #91, opened 2020-12-21
IIGO Implementation for MVP :o

preetl:
# Summary

Trying again...

- Implements role objects: BasePresident, BaseSpeaker and BaseJudge, to meet MVP requirement for compliant islands
- President, Speaker and Judge interfaces to allow overriding by islands Post-MVP
- Extended BaseClient and Client Interface to include 'communication' functions required for IIGO

N.B. IIGO session is conducted sequentially through orchestration.go to allow communication between role objects

Sorry for the mahoosive PR!

## Additional Information

## Test Plan

Unit tests mostly but more testing to come!

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -33,7 +33,6 @@ jobs:
         with:
           golangci_lint_flags: "--disable-all -E exhaustive"
           tool_name: exhaustive 
-          level: warning 
           fail_on_error: true
       - name: exhaustivestruct
         uses: reviewdog/action-golangci-lint@v1
@@ -46,5 +45,4 @@ jobs:
         with:
           golangci_lint_flags: "--disable-all -E goimports"
           tool_name: goimports 
-          level: warning
           fail_on_error: true


QFSW:

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
lhl2617: `.vscode` was already added to `.gitignore` but perhaps too late.

Do:
`git rm --cached -r .vscode`
then commit to remove it

internal/common/baseclient/baseclient.go
```diff
@@ -3,6 +3,7 @@ package baseclient
 
 import (
 	"fmt"
+	"github.com/SOMAS2020/SOMAS2020/internal/common/roles"
```
lhl2617: this should be moved below automatically by `gofmt` in `vscode`--URL imports should be below of intern libs.

.vscode/launch.json
QFSW: You can also do
```
git rm -rf --cached *
git add .
```
to cover _anything_ that was .gitignored
It will look like all you work is gone but when staging the commit everything will be back except the .gitignored stuff

internal/common/gamestate/gamestate.go
```diff
@@ -13,6 +13,9 @@ type GameState struct {
 	// Turn represents the current (1-index) Turn of the game.
 	Turn uint
 
+	// CommonPool represents the amount of ressources in the common pool.
+	CommonPool int
```
QFSW: Perhaps uint if its strictly positive

internal/common/baseclient/baseclient_iigo.go
```diff
@@ -0,0 +1,42 @@
...
+func (c *BaseClient) RuleProposal() string {
+	allRules := rules.AvailableRules
+	for k, _ := range allRules {
+		return k
+	}
+	return ""
+}
+
+// GetClientPresidentPointer is called by IIGO to get the client's implementation of the President Role
+func (c *BaseClient) GetClientPresidentPointer() roles.President {
```
QFSW: More so me needing explanation rather than an issue, but why are the clients responsible for retrieving the president, judge etc. Is this not a server side thing?

internal/common/gamestate/gamestate.go
```diff
@@ -13,6 +13,9 @@ type GameState struct {
 	// Turn represents the current (1-index) Turn of the game.
 	Turn uint
 
+	// CommonPool represents the amount of ressources in the common pool.
```
QFSW: typo

internal/common/shared/common_pool.go
```diff
@@ -0,0 +1,15 @@
+package shared
+
+import "math/rand"
+
+type baseCommonPool struct {
+	amount int
```
QFSW: uint?

internal/server/iigo.go
```diff
@@ -1,10 +1,13 @@
 package server
 
+import "github.com/SOMAS2020/SOMAS2020/internal/server/iigointernal"
+
 // runIIGO : IIGO decides rule changes, elections, sanctions
 func (s *SOMASServer) runIIGO() error {
 	s.logf("start runIIGO")
 	defer s.logf("finish runIIGO")
 	// TODO:- IIGO team
+	_ = iigointernal.RunIIGO(&s.gameState, &s.clientMap)
```
QFSW: What are we discarding?

internal/server/iigointernal/base_judge.go
```diff
@@ -0,0 +1,270 @@
...
+	id                int
+	budget            int
+	presidentSalary   int
+	BallotID          int
+	ResAllocID        int
+	speakerID         int
+	presidentID       int
+	evaluationResults map[int]roles.EvaluationReturn
+	clientJudge       roles.Judge
+}
```
QFSW: Make sure that PascalCase is used for anything that should be exported

internal/server/iigointernal/base_judge.go
```diff
@@ -0,0 +1,270 @@
+package iigointernal
+
+import (
+	"github.com/SOMAS2020/SOMAS2020/internal/common/gamestate"
+	"github.com/SOMAS2020/SOMAS2020/internal/common/roles"
+	"github.com/SOMAS2020/SOMAS2020/internal/common/rules"
+	"github.com/pkg/errors"
+	"math/rand"
```
QFSW: Move this import above the others

internal/server/iigointernal/base_president.go
```diff
@@ -0,0 +1,204 @@
+package iigointernal
+
+import (
+	"github.com/SOMAS2020/SOMAS2020/internal/common/gamestate"
+	"github.com/SOMAS2020/SOMAS2020/internal/common/roles"
+	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
+	"math/rand"
```
QFSW: Move this import above the others

internal/server/iigointernal/base_president.go
```diff
@@ -0,0 +1,204 @@
...
+	id               int
+	clientPresident  roles.President
+	budget           int
+	speakerSalary    int
+	rulesProposals   []string
+	resourceRequests map[int]int
+	//resourceAllocation map[int]int
+	//ruleToVote         string
+	//taxAmountMap       map[int]int
+}
```
QFSW: Make sure that PascalCase is used for anything that should be exported
Don't leave old members commented out

internal/server/iigointernal/base_speaker.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+type baseSpeaker struct {
+	id            int
+	budget        int
+	judgeSalary   int
+	ruleToVote    string
+	votingResult  bool
+	clientSpeaker roles.Speaker
+}
```
QFSW: Make sure that PascalCase is used for anything that should be exported

internal/server/iigointernal/utilities.go
```diff
@@ -0,0 +1,100 @@
...
+func searchForStringInArray(val string, array []string) (int, error) {
+	for i, v := range array {
+		if v == val {
+			return i, nil
+		}
+	}
+	return 0, errors.Errorf("Not found")
+}
+
+type DataPacket struct {
```
QFSW: If you want to use this union-esque approach then it should be a tagged union: i.e, have a field that specifies which of integer text and boolean are actually valid. Ideally this should be avoided though because you're essentially weakening the strong typing

If this isn't a union esque approach, then I think structs with descriptively named values or individually passing the values would be better

But I may be wrong, please feel free to explain if you disagree

internal/server/iigointernal/base_judge.go
```diff
@@ -0,0 +1,270 @@
...
+
+func (j *BaseJudge) sendPresidentSalary() {
+	if j.clientJudge != nil {
+		amount, err := j.clientJudge.PayPresident()
+		if err == nil {
+			featurePresident.budget = amount
+			return
+		}
+	}
+	amount, _ := j.payPresident()
```
QFSW: Don't discard the error, propagate it

internal/server/iigointernal/base_president.go
```diff
@@ -0,0 +1,204 @@
...
+
+// Set allowed resource allocation based on each islands requests
+func (p *basePresident) evaluateAllocationRequests(resourceRequest map[int]int, availCommonPool int) (map[int]int, error) {
+	p.budget -= 10
```
QFSW: What if the budget becomes negative? And why 10?

internal/server/iigointernal/base_speaker.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (s *baseSpeaker) withdrawJudgeSalary(gameState *gamestate.GameState) error {
+	var judgeSalary = int(rules.VariableMap["judgeSalary"].Values[0])
```
QFSW: Best to avoid magic strings like this if we can. Enum?

internal/server/iigointernal/base_judge.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// Pay the president
+func (j *BaseJudge) payPresident() (int, error) {
+	hold := j.presidentSalary
+	j.presidentSalary = 0
+	return hold, nil
+}
+
+func (j *BaseJudge) setSpeakerAndPresidentIDs(speakerId int, presidentId int) {
```
QFSW: Any reason for these to be one function over 2?

internal/server/iigointernal/base_judge.go
```diff
@@ -0,0 +1,270 @@
...
+	j.speakerID = speakerId
+	j.presidentID = presidentId
+}
+
+func (j *BaseJudge) inspectHistoryInternal() {
+	outputMap := map[int]roles.EvaluationReturn{}
+	for _, v := range TurnHistory {
```
QFSW: I'd rename this to `v1` if the others are `v2` and `v3` but something more descriptive would be even better

internal/server/iigointernal/utilities.go
```diff
@@ -0,0 +1,100 @@
...
+func CheckEnoughInCommonPool(value int, gameState *gamestate.GameState) bool {
+	return gameState.CommonPool >= value
+}
+
+func WithdrawFromCommonPool(value int, gameState *gamestate.GameState) error {
+	if CheckEnoughInCommonPool(value, gameState) {
+		gameState.CommonPool -= value
+		return nil
+	} else {
+		return errors.Errorf("Not enough ressources in the common pool to withdraw the amount '%v'", value)
```
QFSW: typo

internal/server/iigointernal/utilities.go
```diff
@@ -0,0 +1,100 @@
...
+	//communication := Communication{
+	//	recipient: recipient,
+	//	sender:    sender,
+	//	data:      data,
+	//}
+	////Send to islands
+	//print(communication) //// Get rid of this
+}
+
+func collapseBoolean(val bool) int {
```
QFSW: Delete?

internal/server/iigointernal/utilities.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func broadcastToAllIslands(sender int, data map[int]DataPacket) {
+	islandsAlive := rules.VariableMap["islands_alive"]
+	for _, v := range islandsAlive.Values {
+		communicateWithIslands(int(v), sender, data)
+	}
+}
+
+func communicateWithIslands(recipient int, sender int, data map[int]DataPacket) {
```
QFSW: Delete?

internal/server/iigointernal/history.go
```diff
@@ -0,0 +1,22 @@
+package iigointernal
+
+import "github.com/SOMAS2020/SOMAS2020/internal/common/rules"
+
+type Accountability struct {
```
QFSW: Make sure that PascalCase is used for anything that should be exported

internal/server/iigointernal/base_speaker.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (s *baseSpeaker) sendJudgeSalary() {
+	if s.clientSpeaker != nil {
+		amount, err := s.clientSpeaker.PayJudge()
+		if err == nil {
+			featureJudge.budget = amount
+			return
+		}
+	}
+	amount, _ := s.payJudge()
```
QFSW: What's being discarded?

.vscode/launch.json
lhl2617: ooh TIL

internal/server/iigointernal/utilities.go
Neelesh99: Soz, work in progress we were waiting for Infra meeting today to ask something but then we found a cleaner way of doing it

internal/common/baseclient/baseclient_iigo.go
Neelesh99: Clients will be allowed to implement some functions (All that are exposed in the Role interfaces) so that they can put some strategy in and influence how IIGO runs, to keep the data-types and stuff standardised we require them to implement the appropriate interface and pass us their custom version which we run the functions from. Note: All exposed functions are Decision functions, Server still actions them so the clients can't mess with anything they aren't meant to. They can only cheat.

internal/server/iigointernal/base_speaker.go
Neelesh99: An error which is there to satisfy the interface function signature but we are 100% sure will always be nil at runtime if the code gets to that line

internal/common/baseclient/baseclient_iigo.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"github.com/SOMAS2020/SOMAS2020/internal/common/roles"
+	"github.com/SOMAS2020/SOMAS2020/internal/common/rules"
+)
+
+// CommonPoolResourceRequest is called by the President in IIGO to
+// request an allocation of resources from the common pool.
+func (c *BaseClient) CommonPoolResourceRequest() int {
+	return 20
```
Darrekt: Some stuff like this throughout the PR is hardcoded, not sure if its intentional, but if its a placeholder maybe add a TODO comment?

internal/common/rules/globalruleregistration.go
```diff
@@ -29,3 +31,63 @@ func registerDemoRule() {
...
+		{
+			name: "check_allocation_rule",
+			reqVar: []string{
+				"island_allocation",
+				"expected_allocation",
+			},
+			v:   []float64{1, -1, 0},
+			aux: []float64{0},
+		},
+	}
```
Darrekt: Just my opinion, but since these are dynamic rules maybe put them in a separate file? It feels like they could be another struct in `config`, so if we decide that the starting rules should be different its easy to change.

internal/common/rules/globalvariableregistration.go
```diff
@@ -30,4 +30,48 @@ var StaticVariables = []VariableValuePair{
 		VariableName: "max_severity_of_sanctions",
 		Values:       []float64{2},
 	},
+	{
+		VariableName: "no_islands_alive",
+		Values:       []float64{6},
```
Darrekt: Maybe take the number of islands in lines like these from a variable instead? @QFSW has also mentioned in another PR that we may want to be able to extend to more than 1 island per team in the future. While unlikely, it's not a very big change that may go a long way :)

internal/common/shared/common_pool.go
```diff
@@ -0,0 +1,15 @@
...
+}
+
+func (cp *baseCommonPool) checkEnoughInCommonPool() int {
+	return rand.Intn(1)
```
Darrekt: Same thing as my comment above about "placeholder" implementations.

Darrekt (commented):
Will continue reviewing later, this is a HUGE pr. I do sympathise that this was a lot of work and the code generally looks to be of good quality other than small things that @QFSW has very sharply noticed, but I think we'd have to call for a PR of this size to be fully reviewed. (or just wait for me to get off my ass and understand everything slowly)

 In future maybe it would be good to split it up into multiple incremental PRs, with the more foundational ones like `iigointernal.go` going first so reviewers can understand the new content, and the changes to the existing structure in a separate one.

internal/server/iigointernal/utilities.go
lhl2617: Agreed.
```
type DataPacketContentType = int

const (
     DataPacketInt DataPacketContentType = iota
     DataPacketString
     DataPacketBool
)

type DataPacket struct {
     T DataPacketContentType 
     Int *int
     Str *string
     Bool *bool
}
```

Pointers or no pointers up to you--pointers have extra safety of `nil` check, and you can (but less encouraged) even get rid of the enum tag and just try to unwrap everything until you have a non-nil ptr.

.vscode/launch.json
preetl: done, thanks guys!

internal/common/gamestate/gamestate.go
preetl: Changed it, good shout!

internal/common/baseclient/baseclient_iigo.go
preetl: Yeah, they are placeholders - i've added in TODOs

internal/common/shared/common_pool.go
preetl: going to delete this file! it was a sandbox - whoops!

internal/common/shared/common_pool.go
preetl: deleting this file - unused!

internal/common/rules/globalruleregistration.go
Neelesh99: That's a good point. Could we target this for post MVP though, just to limit the scope of this PR to iigo?

internal/server/iigointernal/base_president.go
preetl: 10 was an arbitrary choice to make sure that the budget always returns to 0 by the end of an IIGO session. Basically, every 'action' needs a cost but these costs haven't been designed in detail yet. What we have here will never be negative but post MVP we will make a more robust budget deduction system, probably by calling an DeductFromBudget function, external to the object, with the appropriate checks.

internal/common/baseclient/baseclient.go
```diff
@@ -74,3 +86,20 @@ func (c *BaseClient) GameStateUpdate(gameState gamestate.ClientGameState) {
 	c.Logf("Received game state update: %#v", gameState)
 	c.clientGameState = gameState
 }
+
+// Communication is a general datastructure used for communications
+type Communication struct {
+	IntegerData int
+	TextData    string
+	BooleanData bool
+}
```
lhl2617: See comment on tagged union for `DataPacket`.

internal/common/baseclient/baseclient_iigo.go
```diff
@@ -0,0 +1,53 @@
+package baseclient
```
lhl2617: just naming this file `iigo.go` would be fine i reckon

internal/common/roles/judge.go
```diff
@@ -0,0 +1,19 @@
...
+
+type EvaluationReturn struct {
+	Rules       []rules.RuleMatrix
+	Evaluations []bool
+}
+type Judge interface {
+	PayPresident() (int, error)
+	InspectHistory() (map[int]EvaluationReturn, error)
+	DeclareSpeakerPerformance() (int, bool, int, bool, error)
+	DeclarePresidentPerformance() (int, bool, int, bool, error)
```
lhl2617: that's a lotta return types--use a struct? tuple return types are actually not first class citizens in go.

https://stackoverflow.com/questions/13670818/pair-tuple-data-type-in-go

internal/common/roles/president.go
```diff
@@ -0,0 +1,10 @@
+package roles
+
+//President Object
```
lhl2617: good place to give a general overview on how/where this is used.

internal/common/rules/globalvariableregistration.go
lhl2617: go ahead and use `len(shared.TeamIDs)`

internal/server/iigo.go
lhl2617: seems like the `err`, please wrap and return it?

internal/server/iigo.go
```diff
@@ -1,16 +1,65 @@
...
 	return nil
 }
 
 func (s *SOMASServer) runIIGOEndOfTurn() error {
 	s.logf("start runIIGOEndOfTurn")
 	defer s.logf("finish runIIGOEndOfTurn")
-	// TODO:- IIGO team
+	clientMap := s.clientMap
+	for i, v := range clientMap {
+		tax := v.GetTaxContribution()
```
lhl2617: I wouldn't want to pay taxes if I were dead! :P  
Use `getNonDeadClients` :)

internal/server/iigo.go
```diff
@@ -1,16 +1,65 @@
...
+		})
+	}
+	return nil
+}
+
+func (s *SOMASServer) runIIGOAllocations() error {
+	s.logf("start runIIGOAllocations")
+	defer s.logf("finish runIIGOAllocations")
+	clientMap := s.clientMap
+	for i, v := range clientMap {
```
lhl2617: ditto, no allocs if I'm dead!

internal/server/iigointernal/base_judge.go
```diff
@@ -0,0 +1,271 @@
...
+	j.BallotID = 0
+	j.ResAllocID = 0
+}
+
+// Withdraw president salary from the common pool
+// Call common withdraw function with president as parameter
+func (j *BaseJudge) withdrawPresidentSalary(gameState *gamestate.GameState) error {
+	var presidentSalary = int(rules.VariableMap["presidentSalary"].Values[0])
+	var withdrawError = WithdrawFromCommonPool(presidentSalary, gameState)
+	if withdrawError != nil {
```
lhl2617: wait what?????

lhl2617: this means there is an error

internal/server/iigointernal/base_speaker.go
lhl2617: just remove the error in the interface if so :)

internal/server/iigointernal/base_judge.go
```diff
@@ -0,0 +1,271 @@
...
+		featureJudge.presidentSalary = presidentSalary
```
lhl2617: took me a whole min to try to find where `featureJudge` is... maybe let's not mutate package-level globals? I'm definitely not a fan of those! makes code very hard to debug.

internal/server/iigointernal/base_judge.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+	return withdrawError
+}
+
+func (j *BaseJudge) sendPresidentSalary() {
+	if j.clientJudge != nil {
+		amount, err := j.clientJudge.PayPresident()
+		if err == nil {
```
lhl2617, line 44: use `err != nil` pattern instead

internal/server/iigointernal/base_judge.go
```diff
@@ -0,0 +1,271 @@
...
+
+func (j *BaseJudge) inspectHistory() (map[int]roles.EvaluationReturn, error) {
+	j.budget -= 10
+	if j.clientJudge != nil {
+		outputMap, err := j.clientJudge.InspectHistory()
+		if err != nil {
+			j.inspectHistoryInternal()
+		} else {
+			j.evaluationResults = outputMap
+		}
```
lhl2617: need documentation

internal/server/iigointernal/base_judge.go
```diff
@@ -0,0 +1,271 @@
...
+	return j.evaluationResults, nil
+}
+
+func (j *BaseJudge) inspectBallot() (bool, error) {
+	// 1. Evaluate difference between newRules and oldRules to check
+	//    rule changes are in line with ruleToVote in previous ballot
+	// 2. Compare each ballot action adheres to rules in ruleSet matrix
+	j.budget -= 10
+	rulesAffectedBySpeaker := j.evaluationResults[j.speakerID]
+	indexOfBallotRule, err := searchForRule("inspect_ballot_rule", rulesAffectedBySpeaker.Rules)
```
lhl2617: re: magic strings?

internal/server/iigointernal/base_president.go
```diff
@@ -0,0 +1,206 @@
...
+// Send Tax map all the remaining islands
+// Called by orchestration at the end of the turn
+func (p *basePresident) getAllocationRequests(commonPool int) map[int]int {
+	if p.clientPresident != nil {
+		result, error := p.clientPresident.EvaluateAllocationRequests(p.resourceRequests, commonPool)
+		if error == nil {
+			return result
+		}
+	}
+	result, _ := p.evaluateAllocationRequests(p.resourceRequests, commonPool)
```
lhl2617: if a func never returns an error just dont put it :)

internal/server/iigointernal/base_speaker.go
```diff
@@ -0,0 +1,168 @@
...
+				return err
+			}
+		}
+
+	}
+	return nil
+
+}
+
+func (s *baseSpeaker) appointNextJudge() int {
```
lhl2617: should this not return `shared.ClientID`?

internal/server/iigointernal/orchestration.go
```diff
@@ -0,0 +1,144 @@
...
+	SpeakerIDGlobal = featurePresident.appointNextSpeaker()
+	// Get new President ID
+	PresidentIDGlobal = featureJudge.appointNextPresident()
+
+	// Set judgePointer
+	judgePointer = iigoClients[shared.ClientID(JudgeIDGlobal)].GetClientJudgePointer()
+	// Set speakerPointer
+	speakerPointer = iigoClients[shared.ClientID(SpeakerIDGlobal)].GetClientSpeakerPointer()
+	// Set presidentPointer
+	presidentPointer = iigoClients[shared.ClientID(PresidentIDGlobal)].GetClientPresidentPointer()
```
lhl2617: don't cast it like this, it's unsafe. Instead, return `shared.ClientID` from `appoint...`

internal/server/iigointernal/history.go
```diff
@@ -0,0 +1,22 @@
...
+	clientID int
+	pairs    []rules.VariableValuePair
+}
+
+var TurnHistory []Accountability
```
lhl2617: should this not be stored in `gamestate` instead? or is this required to be used?

internal/server/iigointernal/orchestration.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/SOMAS2020/SOMAS2020/internal/common/gamestate"
+	"github.com/SOMAS2020/SOMAS2020/internal/common/roles"
+	"github.com/SOMAS2020/SOMAS2020/internal/common/rules"
+	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
+	"github.com/pkg/errors"
+)
+
+// featureJudge is an instantiation of the Judge interface
+// with both the Base Judge features and a reference to client judges
+var featureJudge = BaseJudge{
```
lhl2617: not a fan of these package level globals.

Cue Tom Clarke `mutables are evil`

Well we still need mutables, but let's instead pass these around. These should be in the `GameState`, passed in an `IIGOInfo` struct variable perhaps. This struct can then be passed around and mutated. 

I strongly prefer returning mutated values and overwriting the old value, e.g.

```
j := s.gameState.IIGOInfo.Judge 
// do whatever u want with j, or even replace it
s.gameState.IIGOInfo.Judge = j
```

internal/server/iigointernal/utilities.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+func withdrawSalary(value int, gameState *gamestate.GameState) (int, error) {
+	return value, WithdrawFromCommonPool(value, gameState)
+}
+
+const (
+	BallotID                 = iota
+	PresidentAllocationCheck = iota
```
lhl2617: no need `= iota` from second line onwards

internal/server/iigointernal/utilities.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func broadcastToAllIslands(sender int, data map[int]DataPacket) {
+	islandsAlive := rules.VariableMap["islands_alive"]
+	for _, v := range islandsAlive.Values {
+		communicateWithIslands(int(v), sender, data)
+	}
+}
+
+func dataPacketToCommunication(d *DataPacket) baseclient.Communication {
```
lhl2617: communication and datapacket are the same thing?

internal/server/iigointernal/utilities_test.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func TestWithdrawFromCommonPoolDeductsValue(t *testing.T) {
+	fakeGameState := gamestate.GameState{CommonPool: 100}
+	valueToWithdraw := 60
+	_ = WithdrawFromCommonPool(valueToWithdraw, &fakeGameState)
+	unexpectedAmountRemaining := fakeGameState.CommonPool != 40
+	if unexpectedAmountRemaining == true {
+		t.Errorf("Not withdrawing resources from CommonPool correctly.")
+	}
```
lhl2617: use a test table to exercise more common pool values and values to withdraw more programatically

internal/server/iigointernal/utilities.go
```diff
@@ -0,0 +1,118 @@
...
+		TextData:    d.textData,
+		BooleanData: d.booleanData,
+	}
+}
+
+func setIIGOClients(clientMap *map[shared.ClientID]baseclient.Client) {
+	iigoClients = *clientMap
+}
+
+func communicateWithIslands(recipient int, sender int, data map[int]DataPacket) {
```
lhl2617: use `shared.ClientID` please?

internal/server/iigointernal/base_president.go
QFSW: Arbitrary choice is fine, but let's make it a named const, or better yet, part of the config
That makes it both consistent throughout the code (no possibility to change one instance but not the other) and self documenting

Post MVP plan sounds good to me

internal/server/iigointernal/base_judge.go
Eirikalb: I second this

internal/server/iigo.go
preetl: Ah yes! I might call this func getAliveClients - glass half full not half empty :D
Edit: ahhh i see why you said non-dead now, there is a critical state too!

preetl: okay to do this in helper.go? I'd rather not have to pass in clientinfos and clientmap though. maybe i can make it a function on the server i.e s.getNonDeadClients() ?

internal/server/iigo.go
preetl: implemented!

internal/server/iigo.go
lhl2617: o nooooooo merge main ples i made it just now sorry :(

internal/server/helper.go
```diff
@@ -40,6 +40,17 @@ func anyClientsAlive(clientInfos map[shared.ClientID]gamestate.ClientInfo) bool
 	}
 	return false
 }
+// getNonDeadClients returns a map of all clients with a non-dead status
+func getNonDeadClients(clientInfos map[shared.ClientID]gamestate.ClientInfo,
```
lhl2617: onoooo soz pls merge main i already implemented it jn

preetl: oh no! sorry

lhl2617:
sneakily updating main to check my lint settings...

internal/server/iigointernal/base_speaker.go
```diff
@@ -0,0 +1,172 @@
...
+	s.ruleToVote = r
+}
+
+//Asks islands to vote on a rule
+//Called by orchestration
+func (s *baseSpeaker) setVotingResult() {
+	if s.clientSpeaker != nil {
+		result, err := s.clientSpeaker.RunVote(s.ruleToVote)
+		if err != nil {
+			s.votingResult, _ = s.runVote(s.ruleToVote)
```
Eirikalb: Same as other error comments, if runVote never returns any error other than nil, then you should just remove it from the interface

internal/common/baseclient/baseclient_iigo.go
Eirikalb: So this is why all of the base_speaker, judge and president implementation are in the server? Could they simply be called speaker, judge and president then, to avoid naming confusion. Considering that the baseClient could/should in theory be completely overwritten but the server side implementation you have written should not?

internal/server/iigointernal/base_speaker.go
```diff
@@ -0,0 +1,172 @@
...
+// Receive a rule to call a vote on
+func (s *baseSpeaker) setRuleToVote(r string) {
+	s.ruleToVote = r
+}
+
+//Asks islands to vote on a rule
+//Called by orchestration
+func (s *baseSpeaker) setVotingResult() {
+	if s.clientSpeaker != nil {
+		result, err := s.clientSpeaker.RunVote(s.ruleToVote)
```
Eirikalb: Is RunVote implemented anywhere at the moment or is this left for team 6?

Eirikalb: Read a bit more and I think I figured it out.. so if you dont have a clientSide implementation of the SpeakerRunVote you will use the unbiased server side implementation? Am I correct in this understanding? This could also have been done using the pattern in baseClient where the baseClient provides a base implementation and the client inheriting from that baseClient. The upside to this is that the childClient will always implement the interface so all clients always implement all three roles. If they want to change the implementation of a method they simply provide the method themselves in the ChildClient and this implementation will override the BaseClient

Eirikalb (commented):
Amazing work alot of really good aspects in here! Only point I would disagree with here is how the client side of things have been structured. I think that all clients should be forced to implement all three roles, but that this implemenation should be offered through baseClient. This way it is for teams later on to change the code as it is way closer to what they will actually have to implement. As it stands now it is really hard to see where client side implementation of the three Roles actually exists because it is hidden in the server-provided methods, which is not necessarily immediately obvious to people trying to familiarize themselves with the codebase.

internal/server/iigointernal/utilities.go
Neelesh99: I've removed the DataPacket type since it was just a duplicate of baseclient.Communications.
We've added the content type fields and modified the instantiation. (Coming in next push)

internal/server/iigointernal/utilities.go
Neelesh99: Yup, sorry no idea why we chose this approach apart for the ensuing hilarity

Neelesh99: Removed now

internal/server/iigointernal/utilities.go
Neelesh99: Thanks, removed in next push

internal/server/iigointernal/utilities.go
akowal3: I had problem with cyclic imports, as datapacket is defined in utilities and receive function in baseclient takes datapacket as input. It could be that the massive refactoring yesterday fixed it, but I didn't notice

internal/server/iigointernal/utilities.go
Neelesh99: Agreed and changing, included in next push

internal/server/iigointernal/utilities_test.go
Neelesh99: Agreed and added

PR #97, opened 2020-12-21
[INFRA/CORE] Buff up `golangci-lint`

lhl2617:
As title. 

Fixed also some caught typos.


QFSW:

QFSW (approved):
Obviously remove the purposeful typo before merging
Is it possible to configure the dictionary at all? If not, I'd say lets keep it for now, but we may need to remove in the future if it gets really annoying

Tangential, but do you think theres an action for enforcing naming conventions? Might be fab if so

lhl2617:
@QFSW i just realised misspell is part of `golangci-lint` and I actually can unlock the floodgates to make `golangci-lint` super strict... experimenting now.

`misspell` is generally quite good with false positives and usually only reports tpyos thtat aer oen charater awya

PR #98, opened 2020-12-22
[INFRA/CORE] Adjust linter

lhl2617:
As title. Appearing as errors is a bit too aggressive. Separate the jobs and make the other stricter linters return warnings.


QFSW:

QFSW (approved):
Can't actually tell _what_ was adjusted about the linter so ideally the description would have stated so ;)

lhl2617:
@QFSW ...exactly nothing.... f**k the docs.

PR #100, opened 2020-12-22
[INFRA/CORE] Make golint super pedantic

lhl2617:
Add a `.golangci.yml` file for `golangci-lint` to use. Set it to
``` 
issues:
  exclude-use-default: false
```
to unlock the pedantic golint floodgates

Ref: https://golangci-lint.run/usage/configuration/

Test Plan:

These weren't visible before:
image

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,2 @@
+issues:
+  exclude-use-default: false


QFSW:

QFSW (approved):
This should be interesting

PR #105, opened 2020-12-22
[IITO] Add initial IITO features

Eirikalb:
# Summary
Adds the basic implementation for giving gifts, with a serverside and client side implementation
## Additional Information
The IITO gift giving sessions works as following: 
- First the server asks each client to request a giftamount
- These giftRequests are then sent to all clients, and they decide how much to offer
   -  TODO: Add checking of resources available to ensure you cannot overextend.
- Third the offers are broadcast and the clients get to accept or reject gifts. 
- Finally the results of the gifting session is broadcast to all clients, although this could be changed to only change the gamestate 

## Test Plan

- Testing will be added shortly in separate PR


QFSW:

internal/common/baseclient/baseclient.go
```diff
@@ -26,6 +26,15 @@ type Client interface {
 	//IIFO: OPTIONAL
 	MakePrediction() (shared.PredictionInfo, error)
 	ReceivePredictions(receivedPredictions shared.PredictionInfoDict) error
+
+	//IITO: COMPULSORY
+	RequestGift() (int, error)
```
QFSW: Do we not want to be able to say how much gift we're requesting?

internal/common/baseclient/iito.go
```diff
@@ -0,0 +1,58 @@
+package baseclient
+
+import "github.com/SOMAS2020/SOMAS2020/internal/common/shared"
+
+// RequestGift allows clients to signalize that they want a gift
+// This information is fed to OfferGifts of all other clients.
+// COMPULSORY, you need to implement this method
+func (c *BaseClient) RequestGift() (int, error) {
+	return 0, nil
```
QFSW: Maybe make the base impl return an error?

internal/server/iito.go
```diff
@@ -15,3 +19,84 @@ func (s *SOMASServer) runIITOEndOfTurn() error {
...
+	err = s.distributeGiftHistory(giftHistoryDict)
+	if err != nil {
+		return err
+	}
+	// Process actions
+	for key, value := range giftHistoryDict {
+		s.logf("Gifts from %s: %v\n", key, value)
+	}
+	defer s.logf("finish runGiftSession")
+	return nil
```
QFSW, line 48: The defer should be at the beginning, after the `s.logf("start runGiftSession")`

QFSW (commented):
Looks mostly good, some minor comments and lint issues

internal/common/baseclient/baseclient.go
Eirikalb: Yes, this is what this function does, the int is currently not bounded, although this could easily be added if wanted

QFSW: Ah my bad, misread it, looks good!

internal/common/baseclient/iito.go
Eirikalb: This should actually interact with the client gamestate I think and return a number. When thinking about this could probably do without errors as it so simple

internal/common/baseclient/baseclient.go
```diff
@@ -26,6 +26,16 @@ type Client interface {
...
+
+	//IITO: COMPULSORY
+	RequestGift() int
+	OfferGifts(giftRequestDict shared.GiftDict) (shared.GiftDict, error)
+	AcceptGifts(receivedGiftDict shared.GiftDict) (shared.GiftInfoDict, error)
+	UpdateGiftInfo(acceptedGifts map[shared.ClientID]shared.GiftInfoDict) error
+
+	//TODO: THESE ARE NOT DONE yet, how do people think we should implement the actual transfer?
+	SendGift(receivingClient shared.ClientID, amount int) error
+	ReceiveGift(sendingClient shared.ClientID, amount int) error
```
lhl2617, line 38: `uint` I guess

on the note of the transfer impl. -- the server acts as a middle man? The server would call a "GetGifts" function that returns a list of `GiftInfo` to other clients.

"ReceiveGift" is good to record who has given the island a gift.

If you want to pass the client pointers around to the other clients that's fine with me--there's two risks:
1. Malicious clients may call functions intended for the server and become an impostor. This can be solved by passing in only an implementation reserved for Inter-client functions. However this is not fool-proof: clients can cast this interface to any type. I'm fine with this, however, if it makes life much easier.
2. No caller identification. Say Team1 has access to Team2's pointer. How does Team2 know Team1 called that function? Trust? An implementation of [MPC](https://en.wikipedia.org/wiki/Secure_multi-party_computation) that would probably be worth the workload of a year-long project?

lhl2617: The server middle-man impl. does mean more work, but it prevents the two problems above. I feel policing at a level of programming implementation is fair (as opposed to policing cheating resources which is part of the game).

internal/common/shared/gifts.go
```diff
@@ -0,0 +1,29 @@
+package shared
+
+// GiftDict is a dictonary of gift
+type GiftDict = map[ClientID]int
+
+// AcceptReason is a just a logical wrapper on the int, a normal int declaration could easily have been used.
+type AcceptReason int
+
+const (
```
lhl2617, line 9: image

sad but this should shut linter up

internal/common/shared/gifts.go
```diff
@@ -0,0 +1,29 @@
...
+	// Accept ...
+	Accept AcceptReason = 0
+	// DeclineDontNeed ...
+	DeclineDontNeed
+	// DeclineDontLikeYou ...
+	DeclineDontLikeYou
```
lhl2617, line 15: image

internal/common/shared/gifts.go
```diff
@@ -0,0 +1,29 @@
...
+// GiftInfoDict key is island index of the island who received your gift
+type GiftInfoDict map[ClientID]GiftInfo
+
+// GiftInfo is a struct containing the information describing a gift
+// exchange between two islands
+type GiftInfo struct {
+	ReceivingTeam  ClientID
+	OfferingTeam   ClientID
+	OfferAmount    int
+	AcceptedAmount int
```
lhl2617: `uint`?

internal/common/baseclient/baseclient.go
lhl2617: A rejection of the gift can be later updated to the offering client

internal/server/iito.go
```diff
@@ -15,3 +23,84 @@ func (s *SOMASServer) runIITOEndOfTurn() error {
...
+	if err != nil {
+		return err
+	}
+	giftHistoryDict, err := s.getGiftAcceptance(giftOffersDict)
+	if err != nil {
+		return err
+	}
+	err = s.distributeGiftHistory(giftHistoryDict)
+	if err != nil {
+		return err
```
lhl2617, line 42: unwrapped errors make me feel uneasy but I trust your functions to return errors that are descriptive enough

(the errors have stack traces, so hopefully they're ok)

image

internal/server/iito.go
```diff
@@ -1,10 +1,18 @@
 package server
 
+import (
```
lhl2617, line 3: I feel we want to ignore dead clients?

lhl2617 (approved):
image

pre-approving, trust you on this.

internal/server/iito.go
Eirikalb: I tried to add it everywhere I could think of but I can see through once again?

internal/server/iito.go
Eirikalb: Ehehe, yeah the plan was basically to add it in for now to have less work later but I can add another PR post MVP to add some more rigourous error handling

internal/common/shared/gifts.go
Eirikalb: I mean these consts are quite selfexplanatory, can add another comment to add their intended use

Eirikalb: These enums are quite self-explanatory but I can add another definition to satisfy the go gods

internal/common/baseclient/baseclient.go
Eirikalb: Should this gate this PR or should I make another PR implementing the transfer of resources?

internal/common/shared/gifts.go
Eirikalb: sure

internal/common/baseclient/baseclient.go
lhl2617: new PR is better I think! Unless you think it won't clutter this one too much :)

internal/common/shared/gifts.go
lhl2617: hail rob pike and ken thompson

internal/server/iito.go
lhl2617: yeha tbh, done is better than perfect for now..

internal/common/shared/gifts.go
```diff
@@ -0,0 +1,29 @@
+package shared
+
+// GiftDict is a dictonary of gift
```
Darrekt, line 3: minor: "dictionary"

internal/server/iito.go
```diff
@@ -15,3 +23,84 @@ func (s *SOMASServer) runIITOEndOfTurn() error {
...
+func (s *SOMASServer) distributeGiftHistory(acceptedGifts map[shared.ClientID]shared.GiftInfoDict) error {
+	//Process acceptedGifts
+	for _, client := range s.clientMap {
+		err := client.UpdateGiftInfo(acceptedGifts)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
Darrekt, line 106: Seems good, but the logic gets a bit messy here with nested maps. Still a great implementation, but I'd feel safer if we had some tests?

Darrekt (approved):
Nice as always, would love to see some tests just for confidence since the nested `map`s can get a bit unintuitive. I think its also not quite immediately clear how `GiftDict` will be used just from its type and name.

internal/server/iito.go
lhl2617: image

Eirikalb: Yeah testing is incoming I promise hahaha

internal/common/baseclient/baseclient.go
Eirikalb: Moved to issue

PR #121, opened 2020-12-23
[INFRA/IIGO] Patch resource withdrawal

Neelesh99:
# Summary

Withdrawing of salaries had some bugs. They are patched now.

## Additional Information

Also documentation of most functions is in place.

## Test Plan

### DIFF
--- a/.github/pull.yml
+++ b/.github/pull.yml
@@ -1,5 +1,0 @@
-version: "1"
-rules:
-  - base: main
-    upstream: SOMAS2020:main    # change `wei` to the owner of upstream repo
-    mergeMethod: merge
--- a/.github/pull.yml
+++ b/.github/pull.yml
@@ -1,0 +1,5 @@
+version: "1"
+rules:
+  - base: main
+    upstream: SOMAS2020:main    # change `wei` to the owner of upstream repo
+    mergeMethod: merge
--- a/internal/common/rules/globalvariableregistration.go
+++ b/internal/common/rules/globalvariableregistration.go
@@ -48,15 +48,15 @@ var StaticVariables = []VariableValuePair{
 	},
 	{
 		VariableName: "speakerSalary",
-		Values:       []float64{50},
+		Values:       []float64{40},
 	},
 	{
 		VariableName: "judgeSalary",
-		Values:       []float64{50},
+		Values:       []float64{60},
 	},
 	{
 		VariableName: "presidentSalary",
-		Values:       []float64{50},
+		Values:       []float64{80},
 	},
 	{
 		VariableName: "expected_tax_contribution",
--- a/internal/server/iigointernal/base_judge.go
+++ b/internal/server/iigointernal/base_judge.go
@@ -28,6 +28,7 @@ func (j *BaseJudge) init() {
 	j.ResAllocID = 0
 }
 
+// returnPresidentSalary returns the salary to the common pool.
 func (j *BaseJudge) returnPresidentSalary() int {
 	x := j.presidentSalary
 	j.presidentSalary = 0
--- a/internal/server/iigointernal/base_judge.go
+++ b/internal/server/iigointernal/base_judge.go
@@ -108,7 +108,7 @@ func (j *BaseJudge) inspectHistoryInternal() {
 // InspectHistory checks all actions that happened in the last turn and audits them.
 // This can be overridden by clients.
 func (j *BaseJudge) InspectHistory() (map[int]roles.EvaluationReturn, error) {
-	j.budget -= 10
+	j.budget -= 10 // will be removed post-MVP
 	if j.clientJudge != nil {
 		outputMap, err := j.clientJudge.InspectHistory()
 		if err != nil {
@@ -127,7 +127,7 @@ func (j *BaseJudge) inspectBallot() (bool, error) {
 	// 1. Evaluate difference between newRules and oldRules to check
 	//    rule changes are in line with RuleToVote in previous ballot
 	// 2. Compare each ballot action adheres to rules in ruleSet matrix
-	j.budget -= 10
+	j.budget -= 10 // will be removed post-MVP
 	rulesAffectedBySpeaker := j.EvaluationResults[j.speakerID]
 	indexOfBallotRule, err := searchForRule("inspect_ballot_rule", rulesAffectedBySpeaker.Rules)
 	if err == nil {
@@ -144,7 +144,7 @@ func (j *BaseJudge) inspectAllocation() (bool, error) {
 	//    in previous resourceAllocation
 	// 2. Compare each resource allocation action adheres to rules in ruleSet
 	//    matrix
-	j.budget -= 10
+	j.budget -= 10 // will be removed post-MVP
 	rulesAffectedByPresident := j.EvaluationResults[j.presidentID]
 	indexOfAllocRule, err := searchForRule("inspect_allocation_rule", rulesAffectedByPresident.Rules)
 	if err == nil {
@@ -177,7 +177,7 @@ func (j *BaseJudge) declareSpeakerPerformanceInternal() (int, bool, int, bool, e
 // DeclareSpeakerPerformance checks how well the speaker did their job
 func (j *BaseJudge) DeclareSpeakerPerformance() (int, bool, int, bool, error) {
 
-	j.budget -= 10
+	j.budget -= 10 // will be removed post-MVP
 	var BID int
 	var result bool
 	var SID int
@@ -208,7 +208,7 @@ func (j *BaseJudge) declareSpeakerPerformanceWrapped() {
 
 // DeclarePresidentPerformance checks how well the president did their job
 func (j *BaseJudge) DeclarePresidentPerformance() (int, bool, int, bool, error) {
-	j.budget -= 10
+	j.budget -= 10 // will be removed post-MVP
 	var RID int
 	var result bool
 	var PID int
@@ -251,7 +251,7 @@ func (j *BaseJudge) declarePresidentPerformanceInternal() (int, bool, int, bool,
 
 // appointNextPresident returns the island ID of the island appointed to be the president in the next turn
 func (j *BaseJudge) appointNextPresident(clientIDs []shared.ClientID) int {
-	j.budget -= 10
+	j.budget -= 10 // will be removed post-MVP
 	var election voting.Election
 	election.ProposeElection(baseclient.President, voting.Plurality)
 	election.OpenBallot(clientIDs)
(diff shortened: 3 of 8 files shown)


QFSW:

.github/pull.yml
```diff
@@ -0,0 +1,5 @@
+version: "1"
```
lhl2617: very non-ideal, but we should fitler this out...
https://stackoverflow.com/questions/28703140/pull-request-ignore-some-file-changes

QFSW (commented):
Possible to get some regression tests for the aforementioned bugs? Other than that and the lint issue, all looks good to me

.github/pull.yml
preetl: removed from this branch, seems like it will have to be manually removed each time :/

Neelesh99:
> Possible to get some regression tests for the aforementioned bugs? Other than that and the lint issue, all looks good to me

Yup added, thanks for the tip

internal/server/iigointernal/base_judge.go
```diff
@@ -51,18 +57,20 @@ func (j *BaseJudge) sendPresidentSalary() {
 	featurePresident.budget = amount
 }
 
-// Pay the president
+// PayPresident pays the president salary.
 func (j *BaseJudge) PayPresident() (int, error) {
 	hold := j.presidentSalary
 	j.presidentSalary = 0
 	return hold, nil
 }
```
Darrekt, line 66: `sendPresidentSalary` seems like just a wrapper for `PayPresident` and doesn't even propagate or resolve its error. Any reason why they need to be 2 functions?

`PayPresident` never returns an error?

`PayPresident` should be camelCased.

internal/server/iigointernal/base_president.go
```diff
@@ -1,9 +1,10 @@
 package iigointernal
 
 import (
-	"github.com/SOMAS2020/SOMAS2020/internal/common/voting"
 	"math/rand"
 
+	"github.com/SOMAS2020/SOMAS2020/internal/common/voting"
+
```
Darrekt: Nice, but why the rogue newline

internal/server/iigointernal/base_speaker.go
```diff
@@ -22,10 +22,16 @@ type baseSpeaker struct {
 	clientSpeaker roles.Speaker
 }
 
+func (s *baseSpeaker) returnJudgeSalary() int {
+	x := s.judgeSalary
+	s.judgeSalary = 0
```
Darrekt, line 29: Same as comment above.

internal/server/iigointernal/base_judge.go
```diff
@@ -28,17 +28,23 @@ func (j *BaseJudge) init() {
 	j.ResAllocID = 0
 }
 
-// Withdraw president salary from the common pool
-// Call common withdraw function with president as parameter
+func (j *BaseJudge) returnPresidentSalary() int {
+	x := j.presidentSalary
+	j.presidentSalary = 0
```
Darrekt, line 34: Why does it reset the salary? From the function name I thought it would just be a pure getter.

internal/server/iigointernal/base_president.go
```diff
@@ -24,6 +25,12 @@ type basePresident struct {
 	//taxAmountMap       map[int]int
 }
 
+func (p *basePresident) returnSpeakerSalary() int {
+	x := p.speakerSalary
+	p.speakerSalary = 0
```
Darrekt, line 30: Same as above, function name doesn't really imply mutation. Explain in documentation above function?

internal/server/iigointernal/base_president.go
```diff
@@ -55,6 +63,7 @@ func (p *basePresident) PickRuleToVote(rulesProposals []string) (string, error)
 	return rulesProposals[rand.Intn(len(rulesProposals))], nil
```
Darrekt, line 64: If this is a placeholder implementation, add a comment above saying it is so we can revisit later.

internal/server/iigointernal/base_president.go
```diff
@@ -44,8 +51,9 @@ func (p *basePresident) EvaluateAllocationRequests(resourceRequest map[int]int,
 	return resourceAllocation, nil
 }
 
-// Chose a rule proposal from all the proposals
-// need to pass in since this is now functional for the sake of client side
+// pickRuleToVote choses a rule proposal from all the proposals made by the islands.
+// It returns an empty string if no rules were proposed, and the rule name (string)
+// of the selected rule otherwise.
 func (p *basePresident) PickRuleToVote(rulesProposals []string) (string, error) {
 	p.budget -= 10
```
Darrekt, line 58: If this is a "service fee", should it be a var above so it can be modified?

internal/server/iigointernal/base_president.go
```diff
@@ -139,6 +151,7 @@ func (p *basePresident) getAllocationRequests(commonPool int) map[int]int {
 	return result
```
Darrekt, line 152: Error discarded in the line above, does it never return an error?

Darrekt (changes requested):
See my discord post:
> From last night we've changed to a  slightly different philosophy for code reviews: we can't check for 100% correctness, because between the 3 of us (@lhl2617 @QFSW and I), it's impossible to know _everything_ about every bit of the spec and how an individual team decided to implement it. So as long as it does not break CI and has its own test cases that pass, we'll let it go and accept that we may have to revisit for a debug later when someone else uses the code.

In this case, some functions seem overly abstracted to me and makes it hard to review which code is being called where. Nonetheless, I trust there is a good reason for it and abstraction usually is not a bad thing.

~~Another potentially major issue is that resources in this file seem to be `int`, where in @JamesTev 's foraging package it's a `float64`. It would be good if this PR could be changed to use a `type resource = int` instead of the hardcoded `int`s so that if a change is necessary it can be done.~~

EDIT: #128 Standardises the type of `Resource` to a `float`, so wait for that to come in, merge it and modify this PR appropriately.

lhl2617:
Btw @Darrekt type Resources float instead--

We changed this in #128 , yet to be merged. We just need to make sure #128 goes in first now

internal/server/iigointernal/base_judge.go
preetl: This is returning the President's salary to the common pool to deal with the case that there aren't enough resources in the common pool to run but the President's salary has already been withdrawn.

internal/server/iigointernal/base_judge.go
preetl: Good question - PayPresident will be overridable by islands (it's in the President Interface in common/roles) but sendPresidentSalary is not. I'll make some changes to the errors in these 2 funcs, not terribly useful as is 👍

internal/server/iigointernal/base_president.go
preetl: eagle eye! cheers!

internal/server/iigointernal/base_president.go
preetl: Sure thing!

internal/server/iigointernal/base_judge.go
preetl: I've added a comment :)

internal/server/iigointernal/base_president.go
preetl: All the budget deductions in this PR for basespeaker, basepresident, and basejudge are placeholders as we haven't designed a proper costing system. Eventually we plan to have different 'service fees' for different actions. If it's alright with you, i'll add a comment but otherwise leave this as is because it will be overhauled soon anyhow.

internal/server/iigointernal/base_president.go
preetl: done!

internal/server/iigointernal/base_president.go
Neelesh99: Yup, the error is there since we need this base implementation to satisfy the interface function signature, but our base implementation itself doesn't return an error ever.

internal/server/iigointernal/base_president.go
Darrekt: `10` seems to be hardcoded in many places, I think it would be good to replace this upfront, especially with #128 coming in which will standardise the type of `Resources`.

internal/server/iigointernal/base_judge.go
Darrekt: Right, I get it. I thought `presidentSalary` was a piece of "information", so it actually holds the value. No problem then.

internal/common/rules/globalvariableregistration.go
```diff
@@ -48,15 +48,15 @@ var StaticVariables = []VariableValuePair{
 	},
 	{
 		VariableName: "speakerSalary",
-		Values:       []float64{50},
+		Values:       []float64{40},
```
lhl2617, line 51: should this be in config? or somewhere defined?

internal/server/iigointernal/utilities.go
```diff
@@ -67,6 +67,10 @@ func WithdrawFromCommonPool(value int, gameState *gamestate.GameState) error {
 	}
 }
 
+func depositIntoCommonPool(value int, state *gamestate.GameState) {
```
lhl2617, line 70: This should be a server method, usually when you feel the need to pass a pointer around consistently to mutate it, it's a good time to ask "should I instead make this a method of an existing struct?"

It's fine for now (if it works), just a note.

Making it a method means all this code has to live in `internal/server/iigo.go` -- that's fine--[big folders with tons of files are a thing in golang](https://golang.org/src/math/big/), it's definitely fine if you have tons of iigoxxx.go files in `internal/server/.`

Making `iigointernal` means you might hit a cyclic dependency problem--likely what you encountered.

Now, the big question is, do you want to continue coding in this pattern, or move out? That depends on how much change will be post-MVP. If you think you can handle the complexity when extensions are needed, and also rewrite functions for the server because you can't include the server package, in addition to making sure everything correct, then go ahead.

lhl2617: all the ints referring to resources in this PR should also be `shared.Resources`.

lhl2617 (changes requested):
+ [ ] Change to `shared.Resources`
+ [ ] make known reply to comment on `utilities.go`.

internal/common/rules/globalvariableregistration.go
Neelesh99: Addressed in PR #138

internal/server/iigointernal/utilities.go
Neelesh99: Hmmm it's a good question you're asking and one of the reasons we had this sort of separate was to organise it a little better with the number of files we had. We've got another PR open (#138 ) that addressed the shared resources issue. Therefore I will close this one.

Neelesh99:
Ok closing this pr since we are opening another PR (#138 ) with structural changes as well as these changes. We address in there the Resources thing. In terms of moving IIGO out, I think we will leave that for now, since most additional functionality is now logic side not requiring more info.

PR #118, opened 2020-12-23
Common-pool

anhnguyen24:

PR #113, opened 2020-12-22
Infra feat disasters

anhnguyen24:
- Fixed previous typos and naming rules
- Add comments
- Add functions to take resource donation from islands and distribute resource to islands (per request)
- Scratch resource distribution after disaster 
image
(updated damage is the leftover damage that islands will have to take on)

PR #124, opened 2020-12-23
Updated: Common Pool

JamesTev:
# Summary

**Note**: this is effectively a duplicate of #118 by @anhnguyen24 but without the whole lot of old commits that seem to have been included accidentally.

Updated version of #113 with conflicts fixed. Revised version of first PR for common pool:

Fixed previous typos and naming rules
Add comments
Add functions to take resource donation from islands and distribute resource to islands (per request)
Scratch resource distribution after disaster

## Additional Information

## Test Plan

Tests to come soon.

### DIFF
--- a/internal/common/config/config.go
+++ b/internal/common/config/config.go
@@ -33,6 +33,9 @@ type Config struct {
 
 	// Wrapped disaster config
 	DisasterConfig DisasterConfig
+
+	// Wrapped commonpool config
+	CommonPoolConfig CommonPoolConfig
 }
 
 // DeerHuntConfig is a subset of foraging config
@@ -63,6 +66,10 @@ type DisasterConfig struct {
 	GlobalProb             float64               // Bernoulli 'p' param. Chance of a disaster occurring
 	SpatialPDFType         shared.SpatialPDFType // Set x,y prob. distribution of the disaster's epicentre (more post MVP)
 	MagnitudeLambda        float64               // Exponential rate param for disaster magnitude
+}
+
+type CommonPoolConfig struct {
+	Resource, Threshold		uint 			
 }
 
 // ForagingConfig captures foraging-specific config
@@ -96,13 +103,17 @@ func GameConfig() Config {
 		FishingConfig:  fishingConf,
 	}
 	disasterConf := DisasterConfig{
-		XMin:            0.0,
-		XMax:            10.0, // chosen quite arbitrarily for now
-		YMin:            0.0,
-		YMax:            10.0,
-		GlobalProb:      0.1,
+		XMin:            				0.0,
+		XMax:            				10.0, // chosen quite arbitrarily for now
+		YMin:            				0.0,
+		YMax:            				10.0,
+		GlobalProb:      				1,
 		SpatialPDFType:  shared.Uniform,
-		MagnitudeLambda: 1.0,
+		MagnitudeLambda: 				1.0,
+	}
+	commonPoolConf := CommonPoolConfig{
+		Resource:				900,		//arbitrarily chosen for initial value
+		Threshold:				1000,		//arbitrarily chosen for initial value
 	}
 
 	return Config{
@@ -114,5 +125,6 @@ func GameConfig() Config {
 		MaxCriticalConsecutiveTurns: 3,
 		ForagingConfig:              foragingConf,
 		DisasterConfig:              disasterConf,
+		CommonPoolConfig:			 commonPoolConf,
 	}
 }
--- a/internal/common/config/config.go
+++ b/internal/common/config/config.go
@@ -103,17 +103,17 @@ func GameConfig() Config {
 		FishingConfig:  fishingConf,
 	}
 	disasterConf := DisasterConfig{
-		XMin:            				0.0,
-		XMax:            				10.0, // chosen quite arbitrarily for now
-		YMin:            				0.0,
-		YMax:            				10.0,
-		GlobalProb:      				1,
+		XMin:            0.0,
+		XMax:            10.0, // chosen quite arbitrarily for now
+		YMin:            0.0,
+		YMax:            10.0,
+		GlobalProb:      1,
 		SpatialPDFType:  shared.Uniform,
-		MagnitudeLambda: 				1.0,
+		MagnitudeLambda: 1.0,
 	}
 	commonPoolConf := CommonPoolConfig{
-		Resource:				900,		//arbitrarily chosen for initial value
-		Threshold:				1000,		//arbitrarily chosen for initial value
+		Resource:		900,		//arbitrarily chosen for initial value
+		Threshold:		1000,		//arbitrarily chosen for initial value
 	}
 
 	return Config{
--- a/internal/common/config/config.go
+++ b/internal/common/config/config.go
@@ -107,7 +107,7 @@ func GameConfig() Config {
 		XMax:            10.0, // chosen quite arbitrarily for now
 		YMin:            0.0,
 		YMax:            10.0,
-		GlobalProb:      1,
+		GlobalProb:      0.1,
 		SpatialPDFType:  shared.Uniform,
 		MagnitudeLambda: 1.0,
 	}
--- a/internal/common/disasters/commonpooldefinition.go
+++ b/internal/common/disasters/commonpooldefinition.go
@@ -1,0 +1,56 @@
+package disasters
+
+import (
+	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
+)
+
+type Commonpool struct {															
+	Resource, Threshold  uint
+}
+
+// DisasterMitigate mitigates the disaster's damage using CP before it hits the island
+// 	+ If cp have enough resource to fully mitigate the disaster, island's personal resources won't be affected 
+//	+ Else, island will be affected with leftover damaged. Each island receives proportional helps from cp with respect to the total damage on 6 islands 
+//	+ EX: if island 1 gets hit the most then cp will help island 1 more than the rest 
+
+//	+ There is a threshold T that incentivizes contributing to the cp from islands
+//	+ If threshold T is met, which mean islands have prepared well before the disaster hit, the damage that disaster have on cp and islands will be half 
+//	+ If not, cp and islands will feel the full effect of disaster
+//	+ For now, each 1 effect of disaster is equivalent to 1000 resource 
+func (e *Environment) DisasterMitigate(individualEffect map[shared.ClientID]float64, proportionalEffect map[shared.ClientID]float64) map[shared.ClientID]float64 {				//input is the porportional effect of each island
+	// compute total effect of disaster on 6 islands 
+	totalEffect := 0.0				
+	updatedIndividualEffect := map[shared.ClientID]float64{}
+
+	for _, effect := range individualEffect {
+		totalEffect = totalEffect + effect
+	}
+
+	// compute cp power towards disaster
+	var newMagnitude float64
+	tempResource := float64(e.CommonPool.Resource)
+	if e.CommonPool.Resource >= e.CommonPool.Threshold { 				//exceeds cp threshold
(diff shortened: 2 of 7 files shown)


internal/common/config/config.go
```diff
@@ -96,13 +103,17 @@ func GameConfig() Config {
...
-		XMin:            0.0,
-		XMax:            10.0, // chosen quite arbitrarily for now
-		YMin:            0.0,
-		YMax:            10.0,
-		GlobalProb:      0.1,
+		XMin:            				0.0,
+		XMax:            				10.0, // chosen quite arbitrarily for now
+		YMin:            				0.0,
+		YMax:            				10.0,
+		GlobalProb:      				1,
```
JamesTev: Not too sure what's happened with the weird spacing here

internal/common/disasters/commonpooldefinition.go
```diff
@@ -0,0 +1,56 @@
...
+// 	+ If cp have enough resource to fully mitigate the disaster, island's personal resources won't be affected 
+//	+ Else, island will be affected with leftover damaged. Each island receives proportional helps from cp with respect to the total damage on 6 islands 
+//	+ EX: if island 1 gets hit the most then cp will help island 1 more than the rest 
+
+//	+ There is a threshold T that incentivizes contributing to the cp from islands
+//	+ If threshold T is met, which mean islands have prepared well before the disaster hit, the damage that disaster have on cp and islands will be half 
+//	+ If not, cp and islands will feel the full effect of disaster
+//	+ For now, each 1 effect of disaster is equivalent to 1000 resource 
+func (e *Environment) DisasterMitigate(individualEffect map[shared.ClientID]float64, proportionalEffect map[shared.ClientID]float64) map[shared.ClientID]float64 {				//input is the porportional effect of each island
+	// compute total effect of disaster on 6 islands
```
JamesTev, line 21: As per previous requests, the server no longer stores an Environment pointer but rather just an instance of the struct itself.

If this is indeed the desired behaviour, we will need to modify the env copy in this function and return the updated env to be reassigned to the game state after I suppose?

internal/common/disasters/environmentdefinition.go
```diff
@@ -58,13 +59,23 @@ func (e Environment) SampleForDisaster() Environment {
...
 	for _, island := range e.Geography.Islands {
 		effect := e.LastDisasterReport.Magnitude / math.Hypot(island.X-epiX, island.Y-epiY) // effect on island i is inverse prop. to square of distance to epicentre
-		out[island.ID] = math.Min(effect, e.LastDisasterReport.Magnitude)                   // to prevent divide by zero -> inf
+		individualEffect[island.ID] = math.Min(effect, e.LastDisasterReport.Magnitude)                   // to prevent divide by zero -> inf
+		totalEffect = totalEffect + individualEffect[island.ID]
 	}
-	return out
+
+	for _, island := range e.Geography.Islands {
+		proportionalEffect[island.ID] =  individualEffect[island.ID] / totalEffect
```
JamesTev, line 77: Surely this doesn't need a separate loop? Could be computed and assigned in previous loop

internal/common/disasters/helpers.go
```diff
@@ -31,17 +31,37 @@ func (report DisasterReport) Display() string {
 }
 
 // DisplayReport is a string format method to viz a disaster report and its effect
-func (env Environment) DisplayReport() string {
+// DisplayReport also visualize how the disaster has been mitigated by the common pool
+func (env *Environment) DisplayReport() (string, map[shared.ClientID]float64) {
 	disasterReport := env.LastDisasterReport.Display()
```
JamesTev, line 36: Same story here with env pointer

internal/common/disasters/setup.go
```diff
@@ -8,13 +8,15 @@ import (
 // InitEnvironment initialises environment according to definitions
 func InitEnvironment(islandIDs []shared.ClientID) Environment {
 	envConf := config.GameConfig().DisasterConfig
-
+	cpConf := config.GameConfig().CommonPoolConfig
+	
+	cp := Commonpool{Resource: cpConf.Resource, Threshold: cpConf.Threshold}
```
JamesTev, line 13: Nice!

internal/common/disasters/environmentdefinition.go
Darrekt: `totalEffect` is accumulated in the first loop, and used as a denominator in the second loop. Fair enough, but it would be more readable with a `.map` style function instead of another naked loop. (If only we weren't in go)

internal/common/disasters/commonpooldefinition.go
Darrekt: Hold up; So is this supposed to actually mutate the environment?

From what I can see, this function is a pure function and computes an update, so there's no reason for it to have a pointer receiver in this case.

internal/common/disasters/helpers.go
```diff
@@ -31,17 +31,37 @@ func (report DisasterReport) Display() string {
...
-	for islandID, effect := range env.DisasterEffects() {
+	sb.WriteString(disasterReport)
+	sb.WriteString("\n------------------------ Disaster Effects ------------------------\n")
+
+	proportionalEffect := map[shared.ClientID]float64{}
+	individualEffect := map[shared.ClientID]float64{}
+	individualEffect, proportionalEffect = env.DisasterEffects()
+
+	for islandID, effect := range individualEffect {
+
```
Darrekt, line 49: Rogue newline

internal/common/disasters/commonpooldefinition.go
lhl2617: L52 mutates it. 

The server's containment of Environment does not matter--the same principle still applies: pointer receiver methods modify the caller, value receiver methods don't.

internal/common/disasters/helpers.go
```diff
@@ -31,17 +31,37 @@ func (report DisasterReport) Display() string {
...
+	damageDifference := map[shared.ClientID]float64{}
+	for islandID, _ := range updatedProportionalEffect {
+		damageDifference[islandID] = individualEffect[islandID] * 1000 - updatedProportionalEffect[islandID]
+	}
+
+	sb.WriteString("\n------------------------ Updated Disaster Effects ------------------------\n")
+	for islandID, effect := range individualEffect {
 		island := env.Geography.Islands[islandID]
-		sb.WriteString(fmt.Sprintf("island ID: %d, \txy co-ords: (%.2f, %.2f), \tdisaster effect: %.2f \n", islandID, island.X, island.Y, effect))
+		sb.WriteString(fmt.Sprintf("island ID: %d, \txy co-ords: (%.2f, %.2f), \tdisaster effect: %.2f, \tUpdated damage: %.2f, \t Common pool mitigated: %.2f \n", islandID, island.X, island.Y, effect, updatedProportionalEffect[islandID], damageDifference[islandID] ))
```
Darrekt: Chonky long line, any way to split the string formatting?

internal/common/disasters/commonpooldefinition.go
lhl2617: I generally _always_ prefer reassignment structs after modifying them (no matter using pointer receiver methods or by getting a new value out of a value receiver), but this method works as well. Up to you.

Note that in Python, it's always implicitly pointer receiver for all objects.

internal/server/commonpool.go
```diff
@@ -0,0 +1,63 @@
...
+	for res := range resChan {
+		id, ci := res.ID, res.Ci
+		s.gameState.ClientInfos[id] = ci
+		
+	}
+}
+
+// islandContribute takes resource donation from individual island into the common-pool
+func (s *SOMASServer) islandContribute(resource int, islandID shared.ClientID) {
+	resChan := make(chan clientInfoUpdateResult, 1)
```
lhl2617: why channel?

lhl2617: I used channels to enable concurrency (something I should not have done, in painful hindsight). Just use normal code?

internal/server/commonpool.go
```diff
@@ -0,0 +1,63 @@
...
+
+	ci := s.gameState.ClientInfos[islandID]
+	ci.Resources += resource 	
+	resChan <- clientInfoUpdateResult{
+		ID:  islandID,
+		Ci:  ci,
+		Err: nil,
+	}			
+	s.gameState.ClientInfos[islandID] = ci
+	s.gameState.Environment.CommonPool.Resource -= uint(resource)
```
Darrekt: Redundant concurrency, the boss will speak with you shortly.

internal/server/commonpool.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+//islandDeplete depletes island's resource based on the severity of the storm
+//Can resource go to negative?
+func (s *SOMASServer) islandDeplete(porpotionEffect map[shared.ClientID]float64) {
```
Darrekt: porpotion

Darrekt (changes requested):
Minor spelling, redundant concurrency? Not the best on concurrency as I've not personally worked with it, @lhl2617 will have more to say on that.

internal/common/disasters/commonpooldefinition.go
JamesTev: @Darrekt yes - that was what I was saying. That it shouldn't be a pointer receiver and we should use reassignment rather. I was just referring to server's containment as an example of previous time this came up.

internal/server/commonpool.go
JamesTev: @anhnguyen24 :) ? Haven't looked at go concurrency at all yet myself

internal/common/disasters/commonpooldefinition.go
Darrekt: _immutable state_

internal/server/commonpool.go
anhnguyen24: I honestly have no idea what went through our brains when we redo this part.. the original code works and I changed it.. Probably got a bit bamboozled :p will redo all the code to related to resource in server. Thanks @lhl2617 @JamesTev !!!

lhl2617: merry somas

internal/common/disasters/helpers.go
lhl2617: image

internal/server/commonpool.go
lhl2617: image

internal/server/commonpool.go
anhnguyen24: image

lhl2617: i cri evrytim

internal/server/commonpool.go
```diff
@@ -3,61 +3,29 @@ import (
...
-	resChan := make(chan clientInfoUpdateResult, 1)
-
 	ci := s.gameState.ClientInfos[islandID]
-	ci.Resources += resource 	
-	resChan <- clientInfoUpdateResult{
-		ID:  islandID,
-		Ci:  ci,
-		Err: nil,
-	}			
+	ci.Resources = ci.Resources + resource
```
NGovani, line 9: This should be logged in the console saying that the client has gained X resources and the common pool has lost X resources. 
Also potentially there should be a function in the client notifying them that they have received resources from the common pool so that they can update their internal state.

internal/server/commonpool.go
```diff
@@ -3,61 +3,29 @@ import (
...
-	close(resChan)
-	for res := range resChan {
-		id, ci := res.ID, res.Ci
-		s.gameState.ClientInfos[id] = ci
-		
-	}
 }
 
 // islandContribute takes resource donation from individual island into the common-pool
 func (s *SOMASServer) islandContribute(resource int, islandID shared.ClientID) {
```
NGovani, line 26: @Neelesh99 
In regards to issue #136, you'll probably end up using this function to when updating the common pool for tax resources, which will also handle reducing the client's resources as well.

There should also be logging and client updates similar to above

PR #129, opened 2020-12-24
[INFRA/CORE] Don't deduct resources below 0 in deductCostOfLiving

lhl2617:
# Summary

Previously it was thought that an island can "owe" resources and have negative resources that they need to recoup.

That idea was trash.

## Test Plan

added unit test.


QFSW:

Darrekt (approved):

PR #128, opened 2020-12-23
Bug fixes/improvements around foraging

mpardalos:
+ [x] Addressed #123
+ [X] Used the typedefs for ClientID and Resources wherever necessary

PR #132, opened 2020-12-24
Add IIFO ForageSharing

E-Mac27:
# Summary
Marked as Draft because it depends on #128.

Server will now ping each client at the start of an IIFO turn if they want to share their most recent foraging attempts. This information will be sent out to whoever they stated. 

Clients will be given the optional choice to implement their client functions (MakeForageInfo() and ReceiveForageInfo(...)). 
- MakeForageInfo() if this is unimplemented, the ShareForageInformation struct will be filled with nil values, meaning nothing will be sent. 
- ReceiveForageInfo(...) does not affect the Clients whether or not its implemented.

## Test Plan

Tested in internal/server/iifo_test.go
Also tested manually by filling out internal/team client.go

### DIFF
--- a/internal/common/shared/forage.go
+++ b/internal/common/shared/forage.go
@@ -1,9 +1,5 @@
 package shared
 
-// ForageContribution is a number between 0 and something that
-// represents how many resources the agents will allocate to foraging
-type ForageContribution = float64
-
 // ForageType selects which resource the agents want to forage in
 type ForageType = int
 
@@ -17,7 +13,7 @@ const (
 // ForageDecision is used to represent a foraging decision made by agents
 type ForageDecision struct {
 	Type         ForageType
-	Contribution ForageContribution
+	Contribution Resources
 }
 
 // ForagingDecisionsDict is a map of clients' foraging decisions
--- a/internal/common/shared/forage.go
+++ b/internal/common/shared/forage.go
@@ -18,3 +18,21 @@ type ForageDecision struct {
 
 // ForagingDecisionsDict is a map of clients' foraging decisions
 type ForagingDecisionsDict = map[ClientID]ForageDecision
+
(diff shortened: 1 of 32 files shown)


QFSW:

mpardalos:
@E-Mac27 you can mark this as ready for review, #128 was merged.

internal/common/shared/forage.go
```diff
@@ -18,3 +18,21 @@ type ForageDecision struct {
...
+	SharedFrom ClientID
+}
+
+// MakeForagingDict is a map of clients' -> foraging decisions, their resource obtained, and which
+// clients to send this information to
+type MakeForagingDict = map[ClientID]ShareForageInfomation
+
+// ReceiveForagingDict is a map of clients -> array of information of other clients
+// foraging decisions, their resource obtained, and which clients to sent this information
+type ReceiveForagingDict = map[ClientID][]ShareForageInfomation
```
lhl2617: Can we name this with `Foragxxx`?

lhl2617: Actually, should we change the naming for foraging stuff to follow this one? I think that's a stretch, but at least the types here are `Foragxxx`

lhl2617: > Actually, should we change the naming for _all_ foraging stuff to follow this one? I think that's a stretch, but at least the types here are `Foragxxx`

E-Mac27: Not entirely too sure what to change MakeForagingDict or ReceiveForagingDict to such that its still readable.

lhl2617: `ForagingOfferDict` and `ForagingReceiptDict`?

E-Mac27: Done

NGovani (approved):
Looks good

NGovani (approved):

Issue #119, opened 2020-12-23
Change AcceptGift function in baseclient


NGovani:
Currently the AcceptGift function in the client allows the agent to change the recipient and offerer id, aswell as the amount offered. This would break the system if a malicious agent changed these values.

https://github.com/SOMAS2020/SOMAS2020/blob/a71dad4f23f4fe110a13bc3d13acdb7d7205e712/internal/common/baseclient/iito.go#L25-L36

lhl2617:
@Darrekt taking over for now

QFSW:
I don't think the client should be responsible for mutating or processing the gift offerings, they also shouldn't be able to see gifts that aren't offered to them. How I would imagine this to would

Introduce a `GiftOffer` struct. This just contains the amount, and the client offering that. Can also include the recipient but redundant

Introduce a `GiftResponse` struct. This contains the amount accepted, and the accepted/rejection reason

The server can still use `acceptedGifts` internally, its a decent way of representing it. `AcceptGifts` though should take (by value) a collection of `GiftOffers`. The client should then go and return a collection of `GiftResponse`s with its responses

The server is then responsible for mutating `acceptedGifts` internally based on the `GiftResponses`
Whoever picks this up let me know if you want to discuss this more, but I think this would be a cleaner way to do it that is simpler for the client and solves the "partial mutation" issue that the current system requires

PR #134, opened 2020-12-24
Fix IITO

Darrekt:
# Summary

IITO currently has some issues regarding struct mutation safety and resources being represented as an `int`. This fix standardises it to the `float64` in `shared`.

## Additional Information

Also fixed a bunch of linter errors.

## Test Plan

### DIFF
--- a/internal/common/config/config.go
+++ b/internal/common/config/config.go
@@ -1,5 +1,6 @@
-// Package config contains configurations of the game.
+// Package config contains types for the configuration of the game.
 // DO NOT depend on other packages outside this folder!
+// Add default values etc. in <root>/params.go
 package config
 
 import "github.com/SOMAS2020/SOMAS2020/internal/common/shared"
@@ -12,13 +13,13 @@ type Config struct {
 	// MaxTurns is the maximum numbers of 1-indexed turns to run the game.
 	MaxTurns uint
 
-	// InitialResources is the default number of resources at the start of the game
+	// InitialResources is the default number of resources at the start of the game.
 	InitialResources shared.Resources
 
 	// CostOfLiving is subtracted from an islands pool before
-	// the next term. This is the simulation-level equivalent to using resources to stay
+	// the next turn. This is the simulation-level equivalent to using resources to stay
 	// alive (e.g. food consumed). These resources are permanently consumed and do
-	// NOT go into the common pool. Note: this is NOT the same as the tax
+	// NOT go into the common pool. Note: this is NOT the same as the tax.
 	CostOfLiving shared.Resources
 
 	// MinimumResourceThreshold is the minimum resources required for an island to not be
@@ -38,13 +39,13 @@ type Config struct {
 // ForagingConfig captures foraging-specific config
 type ForagingConfig struct {
 	// Deer Hunting
-	MaxDeerPerHunt        uint    // Maximimum possible number of deer on a single hunt (regardless of number of participants)
+	MaxDeerPerHunt        uint    // Max possible number of deer on a single hunt (regardless of number of participants)
 	IncrementalInputDecay float64 // Determines decay of incremental input cost of hunting more deer
 	BernoulliProb         float64 // `p` param in D variable (see README). Controls prob of catching a deer or not
 	ExponentialRate       float64 // `lambda` param in W variable (see README). Controls distribution of deer sizes.
 
 	// Deer Population
-	MaxDeerPopulation     uint    // Maximimum possible deer population. Reserved for post-MVP functionality
+	MaxDeerPopulation     uint    // Max possible deer population. Reserved for post-MVP functionality
 	DeerGrowthCoefficient float64 // Scaling parameter used in the population model. Larger coeff => deer pop. regenerates faster
 
 	// TODO: add other pertinent params here (for fishing etc)
@@ -56,38 +57,4 @@ type DisasterConfig struct {
 	GlobalProb             float64               // Bernoulli 'p' param. Chance of a disaster occurring
 	SpatialPDFType         shared.SpatialPDFType // Set x,y prob. distribution of the disaster's epicentre (more post MVP)
 	MagnitudeLambda        float64               // Exponential rate param for disaster magnitude
-}
-
-// GameConfig returns the configuration of the game.
-// (Made a function so it cannot be altered mid-game).
-func GameConfig() Config {
-	foragingConf := ForagingConfig{
-		MaxDeerPerHunt:        4,
-		IncrementalInputDecay: 0.8,
-		BernoulliProb:         0.95,
-		ExponentialRate:       1,
-
-		MaxDeerPopulation:     12,
-		DeerGrowthCoefficient: 0.4,
-	}
-	disasterConf := DisasterConfig{
-		XMin:            0.0,
-		XMax:            10.0, // chosen quite arbitrarily for now
-		YMin:            0.0,
-		YMax:            10.0,
-		GlobalProb:      0.1,
-		SpatialPDFType:  shared.Uniform,
-		MagnitudeLambda: 1.0,
-	}
-
-	return Config{
-		MaxSeasons:                  100,
-		MaxTurns:                    100,
-		InitialResources:            100,
-		CostOfLiving:                10,
-		MinimumResourceThreshold:    5,
-		MaxCriticalConsecutiveTurns: 3,
-		ForagingConfig:              foragingConf,
-		DisasterConfig:              disasterConf,
-	}
 }
--- a/internal/common/disasters/setup.go
+++ b/internal/common/disasters/setup.go
@@ -6,11 +6,9 @@ import (
 )
 
 // InitEnvironment initialises environment according to definitions
-func InitEnvironment(islandIDs []shared.ClientID) Environment {
-	envConf := config.GameConfig().DisasterConfig
-
-	ag := ArchipelagoGeography{Islands: map[shared.ClientID]IslandLocationInfo{}, XMin: envConf.XMin, XMax: envConf.XMax, YMin: envConf.YMin, YMax: envConf.YMin}
-	dp := disasterParameters{globalProb: envConf.GlobalProb, spatialPDF: envConf.SpatialPDFType, magnitudeLambda: envConf.MagnitudeLambda}
+func InitEnvironment(islandIDs []shared.ClientID, disasterConfig config.DisasterConfig) Environment {
+	ag := ArchipelagoGeography{Islands: map[shared.ClientID]IslandLocationInfo{}, XMin: disasterConfig.XMin, XMax: disasterConfig.XMax, YMin: disasterConfig.YMin, YMax: disasterConfig.YMin}
+	dp := disasterParameters{globalProb: disasterConfig.GlobalProb, spatialPDF: disasterConfig.SpatialPDFType, magnitudeLambda: disasterConfig.MagnitudeLambda}
 
 	for i, id := range islandIDs {
 		island := IslandLocationInfo{id, float64(i), float64(0)} // begin with equidistant points on x axis
--- a/internal/common/foraging/deerhunt.go
+++ b/internal/common/foraging/deerhunt.go
@@ -31,9 +31,7 @@ func (d DeerHunt) TotalInput() shared.Resources {
 }
 
 // Hunt returns the utility from a deer hunt
-func (d DeerHunt) Hunt() shared.Resources {
-	fConf := config.GameConfig().ForagingConfig
-
+func (d DeerHunt) Hunt(fConf config.ForagingConfig) shared.Resources {
 	input := d.TotalInput()
 	decay := fConf.IncrementalInputDecay
 	maxDeer := fConf.MaxDeerPerHunt
--- a/internal/common/foraging/deerhunt_test.go
+++ b/internal/common/foraging/deerhunt_test.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"testing"
 
+	"github.com/SOMAS2020/SOMAS2020/internal/common/config"
 	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
 )
 
@@ -37,8 +38,17 @@ func TestDeerUtilityTier(t *testing.T) {
 }
 
 func TestTotalInput(t *testing.T) {
+	fConf := config.ForagingConfig{
+		MaxDeerPerHunt:        4,
+		IncrementalInputDecay: 0.8,
+		BernoulliProb:         0.95,
+		ExponentialRate:       1,
+
+		MaxDeerPopulation:     12,
+		DeerGrowthCoefficient: 0.4,
+	}
 	huntParticipants := map[shared.ClientID]shared.Resources{shared.Team1: 1.0, shared.Team2: 0.9} // arbitrarily chosen for test
-	hunt, _ := CreateDeerHunt(huntParticipants)
+	hunt, _ := CreateDeerHunt(huntParticipants, fConf)
 	ans := hunt.TotalInput()
 	if ans != 1.9 {
 		t.Errorf("TotalInput() = %.2f; want 1.9", ans)
--- a/internal/common/foraging/deerpopulation.go
+++ b/internal/common/foraging/deerpopulation.go
@@ -23,8 +23,7 @@ func (dp *DeerPopulationModel) Logf(format string, a ...interface{}) {
 }
 
 // CreateBasicDeerPopulationModel returns a basic population model based on dP/dt = k(N-y) model. k = growth coeff., N = max deer (constants).
-func createBasicDeerPopulationModel() DeerPopulationModel {
-	fConf := config.GameConfig().ForagingConfig
+func createBasicDeerPopulationModel(fConf config.ForagingConfig) DeerPopulationModel {
 	// definition of deer pop. gradient. Provides dy/dt given y, t.
 	maxDeer := fConf.MaxDeerPopulation
 	deerPopulationGrowth := func(t, y float64) float64 {
--- a/internal/common/foraging/setup.go
+++ b/internal/common/foraging/setup.go
@@ -6,13 +6,12 @@ import (
 )
 
 // CreateDeerHunt receives hunt participants and their contributions and returns a DeerHunt
-func CreateDeerHunt(teamResourceInputs map[shared.ClientID]shared.Resources) (DeerHunt, error) {
-	fConf := config.GameConfig().ForagingConfig
+func CreateDeerHunt(teamResourceInputs map[shared.ClientID]shared.Resources, fConf config.ForagingConfig) (DeerHunt, error) {
 	params := deerHuntParams{p: fConf.BernoulliProb, lam: fConf.ExponentialRate}
 	return DeerHunt{ParticipantContributions: teamResourceInputs, params: params}, nil // returning error too for future use
 }
 
 // CreateDeerPopulationModel returns the target population model. The formulation of this model should be changed here before runtime
-func CreateDeerPopulationModel() DeerPopulationModel {
-	return createBasicDeerPopulationModel()
+func CreateDeerPopulationModel(fConf config.ForagingConfig) DeerPopulationModel {
+	return createBasicDeerPopulationModel(fConf)
 }
--- a/internal/common/shared/disasters.go
+++ b/internal/common/shared/disasters.go
@@ -1,5 +1,12 @@
 package shared
 
+import (
+	"fmt"
+	"log"
+
+	"github.com/SOMAS2020/SOMAS2020/pkg/miscutils"
+)
+
 // Coordinate is a floating point number that can be used represent a position on the map
 type Coordinate = float64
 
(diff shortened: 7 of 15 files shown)


QFSW:

internal/common/baseclient/baseclient.go
```diff
@@ -110,8 +111,10 @@ func (c *BaseClient) GameStateUpdate(gameState gamestate.ClientGameState) {
 	c.clientGameState = gameState
 }
 
-type Role = int
+// Role should be enumerated in some way.
+type Role = int // FIXME
 
+// Enumeration for the different roles.
```
lhl2617: This is fine :D it's how an enum is defined in go

internal/common/baseclient/baseclient.go
```diff
@@ -142,8 +145,10 @@ func (c *BaseClient) GetVoteForElection(roleToElect Role) []shared.ClientID {
 	return returnList
 }
 
-type CommunicationContentType = int
+// CommunicationContentType should be enumerated properly.
+type CommunicationContentType = int // FIXME
 
+// Enumeration type on the different content types.
```
lhl2617: ditto above

internal/common/shared/gifts.go
```diff
@@ -1,7 +1,7 @@
 package shared
 
-// GiftDict is a dictonary of gift
-type GiftDict = map[ClientID]uint
+// GiftDict is a dictionary of gifts that this client is going to receive from other clients
+type GiftDict = map[ClientID]Resources
```
lhl2617: image

internal/server/iito.go
```diff
@@ -48,56 +48,61 @@ func (s *SOMASServer) runGiftSession() error {
...
-	return acceptedGifts, nil
+	return totalResponses, nil
 }
 
-func (s *SOMASServer) distributeGiftHistory(acceptedGifts map[shared.ClientID]shared.GiftInfoDict) error {
-	//Process acceptedGifts
-	for _, client := range s.clientMap {
-		err := client.UpdateGiftInfo(acceptedGifts)
+func (s *SOMASServer) distributeGiftHistory(totalResponses map[shared.ClientID]shared.GiftResponseDict) error {
+	// Process acceptedGifts
```
lhl2617: move comment to above func

internal/server/iito.go
```diff
@@ -48,56 +48,61 @@ func (s *SOMASServer) runGiftSession() error {
...
-			receivedByClientDict[idsend][idto] = giftOffersDict[idto][idsend]
+		offersToThisTeam := shared.GiftOfferDict{}
+		for fromTeam, indivOffers := range totalOffers {
+			offersToThisTeam[fromTeam] = indivOffers[id]
 		}
-	}
-
-	for id, client := range s.clientMap {
-		acceptedGifts[id], err = client.AcceptGifts(receivedByClientDict[id])
+		response, err = client.GetGiftResponses(offersToThisTeam)
```
lhl2617:
```suggestion
		response, err := client.GetGiftResponses(offersToThisTeam)
```

and clear L82-83

internal/server/iito.go
```diff
@@ -48,56 +48,61 @@ func (s *SOMASServer) runGiftSession() error {
...
 	for id, client := range s.clientMap {
-		giftOfferDict[id], err = client.OfferGifts(giftRequestDict)
+		var offer map[shared.ClientID]shared.GiftOffer
+		var err error
+		// Gather all the requests made to this team
+		requestsToThisTeam := shared.GiftRequestDict{}
+		for fromTeam, indivRequests := range totalRequests {
+			requestsToThisTeam[fromTeam] = indivRequests[id]
+		}
+		offer, err = client.GetGiftOffers(requestsToThisTeam)
```
lhl2617:
```suggestion
		offer, err := client.GetGiftOffers(requestsToThisTeam)
```

and remove L65-66

internal/server/iito.go
```diff
@@ -48,56 +48,61 @@ func (s *SOMASServer) runGiftSession() error {
...
-		giftRequestDict[id] = client.RequestGift()
+		giftRequestDict[id] = client.GetGiftRequests()
 	}
 	return giftRequestDict
 }
-func (s *SOMASServer) getGiftOffers(giftRequestDict shared.GiftDict) (map[shared.ClientID]shared.GiftDict, error) {
-	giftOfferDict := map[shared.ClientID]shared.GiftDict{}
-	var err error
+
+// GetGiftOffers collects all responses from clients to their requests in a map
```
lhl2617:
```suggestion
// getGiftOffers collects all responses from clients to their requests in a map
```

internal/common/baseclient/iito.go
```diff
@@ -2,43 +2,53 @@ package baseclient
 
 import "github.com/SOMAS2020/SOMAS2020/internal/common/shared"
 
-// RequestGift allows clients to signalize that they want a gift
+// FIXME: Wait for implementation of getClientsAlive()
+const aliveClients = 6
```
lhl2617: we don't need this, we can just say 
`requests := shared.GiftRequestDict{}` in L12 & L27 -- maps can be dynamic

internal/common/shared/gifts.go
```diff
@@ -1,7 +1,28 @@
...
+// A layer of redundancy exists, in that the key and the OfferingTeam encode the same information,
+// but this may prove useful in implementation.
+type GiftRequestDict = map[ClientID]GiftRequest
+
+// GiftOffer contains the details of a gift offer from an island to another
+type GiftOffer struct {
+	ReceivingTeam ClientID
+	OfferingTeam  ClientID
+	OfferAmount   Resources
+}
```
mpardalos: Could these 2 maybe be merge into just `type Gift struct`?

internal/server/iito.go
```diff
@@ -48,56 +48,61 @@ func (s *SOMASServer) runGiftSession() error {
...
+	totalOffers := map[shared.ClientID]shared.GiftOfferDict{}
+	// Loop over each team
 	for id, client := range s.clientMap {
-		giftOfferDict[id], err = client.OfferGifts(giftRequestDict)
+		var offer map[shared.ClientID]shared.GiftOffer
+		var err error
+		// Gather all the requests made to this team
+		requestsToThisTeam := shared.GiftRequestDict{}
+		for fromTeam, indivRequests := range totalRequests {
+			requestsToThisTeam[fromTeam] = indivRequests[id]
```
NGovani, line 70: Will break if indivRequests does not have a value for the corresponding id
i.e. If team 1 doesn't request from team 6, this will break when team 6's requests are being packaged

NGovani: This will break if indivRequests does not have a value for the corresponding id.
I.e. if team 1 does not request anything from team 2, when team 2's requests are being packaged it will try to find team 1's request for team 2 and then break

A similar issues is found on line 84 and 99

internal/common/baseclient/iito.go
```diff
@@ -2,43 +2,53 @@ package baseclient
...
-// COMPULSORY, you need to implement this method
-func (c *BaseClient) OfferGifts(giftRequestDict shared.GiftDict) (shared.GiftDict, error) {
-	return giftRequestDict, nil
+// GetGiftOffers allows clients to make offers in response to gift requests by other clients.
+// It can offer multiple partial gifts.
+// COMPULSORY, you need to implement this method. This placeholder implementation offers no gifts.
+func (c *BaseClient) GetGiftOffers(receivedRequests shared.GiftRequestDict) (shared.GiftOfferDict, error) {
+	offers := make(shared.GiftOfferDict, aliveClients)
+	for key := range offers {
+		offers[key] = shared.GiftOffer{ReceivingTeam: c.GetID(), OfferingTeam: key, OfferAmount: 0.0}
```
NGovani: Receiving team should be the key, while offering team should be the id of the current client.

However clients should not be allowed to specify the offering team since they could pretend to be someone else.

internal/common/shared/gifts.go
Darrekt: Shit lol I didn't even realise how similar they were until now. I'll leave them as-is for now, we might need them to be separate when we make sure all edge cases are covered. (which is probably why I thought we needed 2 in the first place).

internal/common/baseclient/iito.go
```diff
@@ -2,43 +2,53 @@ package baseclient
...
 // This information is fed to OfferGifts of all other clients.
 // COMPULSORY, you need to implement this method
-func (c *BaseClient) RequestGift() uint {
-	if c.clientGameState.ClientInfo.LifeStatus == shared.Critical {
-		return 100
+func (c *BaseClient) GetGiftRequests() shared.GiftRequestDict {
+	requests := make(shared.GiftRequestDict, aliveClients)
+	for key := range requests {
+		if c.clientGameState.ClientInfo.LifeStatus == shared.Critical {
+			requests[key] = shared.GiftRequest{RequestingTeam: c.GetID(), OfferingTeam: key, RequestAmount: 100.0}
```
NGovani: Clients should not be allowed to specify the requesting team since they could pretend to be someone else.

internal/common/baseclient/iito.go
```diff
@@ -2,43 +2,50 @@ package baseclient
...
-func (c *BaseClient) AcceptGifts(receivedGiftDict shared.GiftDict) (shared.GiftInfoDict, error) {
-	acceptedGifts := shared.GiftInfoDict{}
-	for client, offer := range receivedGiftDict {
-		acceptedGifts[client] = shared.GiftInfo{
-			ReceivingTeam:  client,
-			OfferingTeam:   c.GetID(),
-			OfferAmount:    offer,
-			AcceptedAmount: offer,
-			Reason:         shared.Accept}
+func (c *BaseClient) GetGiftResponses(receivedOffers shared.GiftOfferDict) (shared.GiftResponseDict, error) {
```
NGovani: good resolves issue #119

NGovani (changes requested):
some edge case issues

internal/server/iito.go
```diff
@@ -48,56 +48,57 @@ func (s *SOMASServer) runGiftSession() error {
 	return nil
 }
 
-func (s *SOMASServer) getGiftRequests() shared.GiftDict {
-	giftRequestDict := shared.GiftDict{}
+// GetGiftRequests collects all gift requests from the clients in a map
+func (s *SOMASServer) getGiftRequests() map[shared.ClientID]shared.GiftRequestDict {
+	giftRequestDict := map[shared.ClientID]shared.GiftRequestDict{}
 	for id, client := range s.clientMap {
```
E-Mac27: Should this call functions from dead clients as well? 

This can be said for all the rest of the functions that cycles through s.clientMap.

E-Mac27: If not have a look at server/iifo.go. They call getNonDeadClientIDs which returns nonDeadClients.

E-Mac27 (changes requested):

internal/common/baseclient/iito.go
```diff
@@ -2,43 +2,50 @@ package baseclient
...
 
-// RequestGift allows clients to signalize that they want a gift
+// GetGiftRequests allows clients to signalize that they want a gift
 // This information is fed to OfferGifts of all other clients.
 // COMPULSORY, you need to implement this method
-func (c *BaseClient) RequestGift() uint {
-	if c.clientGameState.ClientInfo.LifeStatus == shared.Critical {
-		return 100
+func (c *BaseClient) GetGiftRequests() shared.GiftRequestDict {
+	requests := shared.GiftRequestDict{}
```
E-Mac27: Does this do anything? You seem to be creating an empty dict, then immediately loop through it right after. Should it do anything (considering its a BaseClient implementation)?

This is the same for below GetGiftOffers()

internal/common/baseclient/iito.go
E-Mac27: +1, The server should definitely not take this value into consideration and should override it.

internal/common/baseclient/iito.go
E-Mac27: @NGovani mentioned that this might be cause you were unable to populate this field. If that's the case, please add a TODO so others know.

internal/server/iito.go
```diff
@@ -48,56 +48,57 @@ func (s *SOMASServer) runGiftSession() error {
...
-	for _, client := range s.clientMap {
-		err := client.UpdateGiftInfo(acceptedGifts)
+// distributeGiftHistory collates all responses to a single client and calls that client to receive its responses.
+func (s *SOMASServer) distributeGiftHistory(totalResponses map[shared.ClientID]shared.GiftResponseDict) error {
+	for id, client := range s.clientMap {
+		responsesToThisTeam := shared.GiftResponseDict{}
+		for fromTeam, indivResponses := range totalResponses {
+			responsesToThisTeam[fromTeam] = indivResponses[id]
+		}
+		err := client.UpdateGiftInfo(responsesToThisTeam)
```
E-Mac27: Should clients be able to send offers to dead clients?

internal/server/iito.go
```diff
@@ -48,56 +48,57 @@ func (s *SOMASServer) runGiftSession() error {
...
-		acceptedGifts[id], err = client.AcceptGifts(receivedByClientDict[id])
+		offersToThisTeam := shared.GiftOfferDict{}
+		for fromTeam, indivOffers := range totalOffers {
+			offersToThisTeam[fromTeam] = indivOffers[id]
+		}
+		response, err := client.GetGiftResponses(offersToThisTeam)
 		if err != nil {
-			return acceptedGifts, err
+			totalResponses[id] = response
 		}
```
E-Mac27: Copied & Pasted statement.
I feel like you got the logic wrong here. This if statement should only be used when an error occurs. Currently the logic is - if there is an error, totalResponses is assigned.   
It might be because it was very late when you did this but lol.

internal/server/iito.go
```diff
@@ -48,56 +48,57 @@ func (s *SOMASServer) runGiftSession() error {
...
+		// Gather all the requests made to this team
+		requestsToThisTeam := shared.GiftRequestDict{}
+		for fromTeam, indivRequests := range totalRequests {
+			requestsToThisTeam[fromTeam] = indivRequests[id]
+		}
+		offer, err := client.GetGiftOffers(requestsToThisTeam)
 		if err != nil {
-			return giftOfferDict, err
+			totalOffers[id] = offer
 		}
```
E-Mac27: I feel like you got the logic wrong here. This if statement should only be used when an error occurs. Currently the logic is - if there is an error, totalResponses is assigned.   
It might be because it was very late when you did this but lol.

internal/common/shared/gifts.go
```diff
@@ -1,7 +1,28 @@
 package shared
 
-// GiftDict is a dictonary of gift
-type GiftDict = map[ClientID]uint
+// GiftRequest contains the details of a gift request from an island to another
```
E-Mac27: Me being a dumb dumb in English got confused with the naming scheme. I had to ask for clarification from the rest of Team1. Is it okay to be more explicit in the comments (same for the below)? 

Moreover, OfferingTeam is used again in GiftOffer - does it still hold the same meaning?

internal/common/baseclient/iito.go
```diff
@@ -47,13 +54,13 @@ func (c *BaseClient) UpdateGiftInfo(acceptedGifts map[shared.ClientID]shared.Gif
 // SendGift is executed at the end of each turn and allows clients to
 // send the gifts promised in the IITO
 // COMPULSORY, you need to implement this method
-func (c *BaseClient) SendGift(receivingClient shared.ClientID, amount int) error {
+func (c *BaseClient) SendGift(offer shared.GiftOffer) error {
 	return nil
 }
```
E-Mac27: Is this meant to be nil?
- If it is meant to be nil, can you include this in the comments
- If not - Is it possible to add a TODO?

Same for below

E-Mac27 (changes requested):

internal/server/iito.go
Darrekt: yeah will flip the `!=` to `==` I was intending to write tests today, I just wanted to put the PR up first so people could comment bc I did this commit all in 1 go and maps on maps in my head usually do not end well without tests

internal/server/iito.go
Darrekt: We're awaiting a change from @mpardalos and @lhl2617 for a getNonDeadClientIDs from the server for clients to use. Once that's implemented I'll put that in throughout this PR. Everything until then is a bit of a placeholder

internal/common/baseclient/iito.go
Darrekt: this is the same as `make(shared.GiftRequestDict)`. It's just initialising a variable containing a map. I originally had it to `make(shared.GiftRequestDict, numOfAliveClients)` to set an appropriate size map containing "zeroed" elements. This implementation will be subject to change once we implement `getNonDeadClientIDs` for clients, I should have left a FIXME on this.

internal/common/baseclient/iito.go
Darrekt: This whole file contains placeholder implementations. The appropriate functions are marked COMPULSORY, which means the implementation we put in here are "useless". In the previous functions we put something down so that agent implementers can read it and have a good idea of what the function is supposed to do, you'll notice that the implementations thus far just had (mostly) the correct looping and data structure logic; they're not supposed to do anything "meaningful" from a gameplay standpoint.

Therefore in this case the only thing we really can do is just return no error and do nothing with the gift offer we received. Open to change of course.

internal/common/shared/gifts.go
Darrekt: as @mpardalos pointed out, the 2 data types are pretty similar for now, they may not have to be separate. They just are for the time being until I make test cases to execute the code, and we know that they don't have to be different for data protection purposes.

internal/server/iito.go
Darrekt: Yuor'e corect.

As @QFSW and I discussed, we should make data types to represent this outcome, and let the server do a "validation" of sorts to make sure data is in the right shape at each phase. This also applies to fixing the other issue you mentioned of clients mutating data inappropriately.

The reason everything is done in atomic map operations in this first draft is because we want clients to be able to make all their decisions at once, instead of being asked for a response on an individual basis: A client who does not like another may still accept a gift from it if it is the only gift it received and it needs resources.

internal/server/iito.go
Darrekt: yep need to flip the sign ty

internal/server/iito.go
Darrekt: Nope! Again, awaiting the exposure of a getNonDeadClients equivalent from the server on another PR.

internal/common/baseclient/iito.go
Darrekt: I wish we were in typescript here, the client could then return a `Partial<GiftOffer>` which omits the `OfferingTeam` field and the server can fill it in with a `.map`. An equivalent Go implementation involves making a whole new type to represent a partial GiftOffer. Any way to do this better @lhl2617? If not I'll go ahead and do it

internal/server/iito.go
```diff
@@ -48,54 +48,61 @@ func (s *SOMASServer) runGiftSession() error {
...
+		requestsToThisTeam := []shared.GiftRequest{}
 		for fromTeam, indivRequests := range totalRequests {
+			// TODO: Will fail if the list does not have all 6 entries. Fix in a sanitisation function.
 			requestsToThisTeam[fromTeam] = indivRequests[id]
 		}
+
 		offer, err := client.GetGiftOffers(requestsToThisTeam)
-		if err != nil {
-			totalOffers[id] = offer
+		if err == nil {
```
NGovani: Wouldn't this mean there is no error, causing the function to return and asking no other clients for their offers?

internal/server/iito.go
```diff
@@ -48,54 +48,61 @@ func (s *SOMASServer) runGiftSession() error {
...
+// getGiftOffers collects all responses from clients to their requests in a map
+func (s *SOMASServer) getGiftOffers(totalRequests map[shared.ClientID][]shared.GiftRequest) (map[shared.ClientID][]shared.GiftOffer, error) {
+	totalOffers := map[shared.ClientID][]shared.GiftOffer{}
 	for id, client := range s.clientMap {
 		// Gather all the requests made to this team
-		requestsToThisTeam := shared.GiftRequestDict{}
+		requestsToThisTeam := []shared.GiftRequest{}
 		for fromTeam, indivRequests := range totalRequests {
+			// TODO: Will fail if the list does not have all 6 entries. Fix in a sanitisation function.
 			requestsToThisTeam[fromTeam] = indivRequests[id]
```
NGovani: Does Go allow an empty list to be accessed like this? If so then its fine, otherwise we need to initialise it with size = number of clients.

Similar issue on Ln 88 & 106

NGovani (changes requested):
Minor changes
Looks good otherwise

internal/server/iito.go
E-Mac27: Okay. I'm resolving conversation and will wait till everything is finished.

internal/common/baseclient/iito.go
E-Mac27: Will there also be server side checks or will we rely on clients using this function?

internal/server/iito.go
E-Mac27: Sounds good.

internal/common/shared/gifts.go
E-Mac27: Okay. I will wait for this to be completed. Resolving convo for now.

internal/common/baseclient/baseclient.go
```diff
@@ -3,9 +3,10 @@ package baseclient
 
 import (
 	"fmt"
-	"github.com/SOMAS2020/SOMAS2020/internal/common/rules"
 	"log"
 
+	"github.com/SOMAS2020/SOMAS2020/internal/common/rules"
+
```
lhl2617:
```suggestion
```

lhl2617:
Closing this. I accidentally force-pushed over it. Your new PR is #145

PR #137, opened 2020-12-24
[INFRA/CORE] Make foraging update resources, notify agents of their foraging returns

mpardalos:
## Summary
This PR makes 2 changes:
+ Makes foraging actually update the clients' resources; first taking away the contributed resources and then giving back the foraging results.
+ Adds a `ForageUpdate` function on the Client that gets called when foraging is finished to with the foraging returns

This PR makes an implicit design decision:
*Foraging rewards are distributed proportional to participants' contributions*. It could easily be changed to be an equal split, or some other method.

### DIFF
--- a/internal/common/baseclient/baseclient.go
+++ b/internal/common/baseclient/baseclient.go
@@ -42,6 +42,7 @@ type Client interface {
 
 	//Foraging
 	DecideForage() (shared.ForageDecision, error)
+	ForageUpdate(forageReturn int)
 
 	//IITO: COMPULSORY
 	RequestGift() uint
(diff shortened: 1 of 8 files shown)


QFSW:

NGovani (approved):
These changes work 
Note for @JamesTev the foraging.go file has been changed and will likely cause some merge conflicts with your fish forage implementation. 
These conflicts shouldn't be major, they just serve as a guideline for how to update the actual gamestate. specifically lines 67-74 in internal/server/forage.go

Sidenote: ForageUpdate notifies the client about their returns from foraging

Issue #135, opened 2020-12-24
!BREAKING CHANGE! Passing server pointer now with a `ReadOnlyServer` Interface 


lhl2617:
This lets clients get the latest game state without complicated game state updates.

The reason we didn't want to do this in the beginning is the risk of people casting the server pointer and doing malicious actions against it.

Unfortunately, this server-client paradigm is _extremely hard_ to wrap our heads around, proven by the sheer amount of broken stuff written to follow this paradigm.

Team 1 is now working on this.

lhl2617:
assigned to @mpardalos

lhl2617:
CC all infra reps, after this resolving, server DOES NOT NEED TO pass newest game state or update game state to clients--clients can call a function on their own copy of the `ReadOnlyServer` pointer, given at the start of the turn.

# THIS ALSO MEANS WE NEED TO MAKE SURE NO AGENT CODE DOES MALICIOUS ACTIONS ON THE SERVER POINTER. FOR THE SAKE OF SECURITY, I WILL NOT GIVE AN EXAMPLE ON HOW THIS CAN BE BROKEN.

# IF YOU SEE ANY AGENT CODE TRYING TO DO ANYTHING SPECIAL TO THE POINTER TO THE SERVER INTERFACE, PLEASE FLAG IT.

 @Darrekt @QFSW @JamesTev @Eirikalb @Diregie-J @Neelesh99 @777jameskim @P-A-Ha

lhl2617:
@mpardalos 
+ [ ] Remove `StartOfTurnUpdate` and `GameStateUpdate`
+ [ ] Make `ReadOnlyServer` interface
+ [ ] Pass pointer to `ReadOnlyServer` to client on startup

lhl2617:
Resolved by #139

PR #139, opened 2020-12-24
Replace GameState updates with read-only reference to server

mpardalos:
----

# Summary

Addresses #135

Based on #137

+ [X] Make `ReadOnlyServer` interface
+ [X] Pass pointer to `ReadOnlyServer` to client on startup
+ [X] Define a function on `ReadOnlyInterface` to get `ClientGameState`
+ [x] Remove `StartOfTurnUpdate` and `GameStateUpdate`
+ [x] Add alive status of other clients to ClientGameState

## Test Plan

- No tests needed/possible


QFSW:
Dont we still need a `StartOfTurn` event in case agents want to inject their own behaviour or decision making / logging at this point?

mpardalos:
Probably a good idea. Just should not have any args.

internal/common/baseclient/baseclient.go
```diff
@@ -89,6 +99,10 @@ func (c *BaseClient) GetID() shared.ClientID {
 	return c.id
 }
 
+func (c BaseClient) Initialise(ServerReadHandle) {}
```
lhl2617: listen to linter

internal/server/iigo.go
```diff
@@ -22,10 +22,13 @@ func (s *SOMASServer) runIIGOEndOfTurn() error {
...
-		s.gameState.CommonPool += tax
-		newGameState := s.gameState.GetClientGameStateCopy(clientID)
-		newGameState.ClientInfo.Resources -= tax
-		v.GameStateUpdate(newGameState)
+
+		err := s.takeResources(clientID, tax, "tax")
+		if err != nil {
+			s.gameState.CommonPool += tax
+		} else {
+			s.logf("Error getting tax from %v: %v", clientID, err)
```
lhl2617, line 31: do we just log n ignore?

internal/server/server.go
```diff
@@ -41,17 +42,27 @@ func SOMASServerFactory() Server {
...
 			ClientInfos:    clientInfos,
 			Environment:    disasters.InitEnvironment(clientIDs),
 			DeerPopulation: foraging.CreateDeerPopulationModel(),
 		},
 	}
+
+	for _, client := range clientMap {
+		client.Initialise(ServerForClient{
+			clientID: client.GetID(),
+			server: server,
```
lhl2617: image

internal/server/server.go
```diff
@@ -41,17 +42,27 @@ func SOMASServerFactory() Server {
...
+		})
+	}
+
+	return server
+
```
lhl2617:
```suggestion
```

lhl2617 (changes requested):

internal/common/gamestate/gamestate.go
```diff
@@ -48,10 +48,16 @@ func (g GameState) Copy() GameState {
 
 // GetClientGameStateCopy returns the ClientGameState for the client having the id.
 func (g *GameState) GetClientGameStateCopy(id shared.ClientID) ClientGameState {
+	clientLifeStatuses := map[shared.ClientID]shared.ClientLifeStatus{}
```
lhl2617, line 51: test this

internal/server/iigo.go
```diff
@@ -22,10 +22,13 @@ func (s *SOMASServer) runIIGOEndOfTurn() error {
 	clientMap := getNonDeadClients(s.gameState.ClientInfos, s.clientMap)
 	for clientID, v := range clientMap {
 		tax := v.GetTaxContribution()
-		s.gameState.CommonPool += tax
-		newGameState := s.gameState.GetClientGameStateCopy(clientID)
-		newGameState.ClientInfo.Resources -= tax
-		v.GameStateUpdate(newGameState)
+
+		err := s.takeResources(clientID, tax, "tax")
```
NGovani, line 26: Potentially need to update the client through calling a function similar to ForageUpdate(), maybe called TaxUpdate(). This would tell the client how much tax was paid, and prompt them to update their gamestate

internal/server/iigo.go
NGovani: That should be fine, we just assume the client has made an invalid move and leave it alone

QFSW:
> Probably a good idea. Just should not have any args.

Agreed

internal/common/baseclient/baseclient.go
mpardalos: I don't see any linter messages for this line

internal/server/iigo.go
mpardalos: Yeah I think this is the right way too.

internal/server/iigo.go
mpardalos: Added

internal/common/gamestate/gamestate.go
mpardalos: Done

mpardalos:
This should be ready to merge now.

@lhl2617 @QFSW

internal/common/baseclient/baseclient.go
```diff
@@ -93,6 +101,10 @@ func (c *BaseClient) GetID() shared.ClientID {
...
```
lhl2617: we should impleemnt this for the baseclient?

and add a variable for the BaseClient struct for it.

Use pointer rcvers btw

lhl2617 (changes requested):
+ [ ] impl baseclient
+ [ ] optional... fix documentation issues for linter

internal/common/baseclient/baseclient.go
mpardalos: Is there any reason to keep that in the base client? I could add it but it won't be used. 
I thought baseclient is supposed to have "null" implementations.

lhl2617: nope, baseclient is supposed to implement a working client as a basis for extensions

mpardalos: ptr receivers done

mpardalos: > nope, baseclient is supposed to implement a working client as a basis for extensions

Makes sense. Keeping the ServerReadHandle in the baseclient.

lhl2617 (approved):
image

PR #144, opened 2020-12-27
[INFRA/CORE] Move logger to pkg

lhl2617:
As title.

More logical area to put it as it is not an internal SOMAS pkg.

## Test Plan
Everything passes as usual.


QFSW:

QFSW (approved):

PR #147, opened 2020-12-27
Implement Borda Count in voting

Ning-72:


QFSW:

QFSW (changes requested):
Many of the same issues as #86, no description, messy for loop syntax
Please make the changes that we suggest

Diregie-J:
> 与[＃86](https://github.com/SOMAS2020/SOMAS2020/pull/86)相同的许多问题，没有描述，循环语法混乱，
> 请进行建议的更改

Changes required have been made.

Ning-72:
> Many of the same issues as #86, no description, messy for loop syntax
> Please make the changes that we suggest

Changes required have been made.

internal/common/voting/election.go
```diff
@@ -55,8 +55,90 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+	var votesLayoutElect map[int][]int
+	votesSliceSquare := e.votes
+	candidatesNumber = len(e.islandsToVote)
+	islandsNumber = len(votesSliceSquare)
+	scoreInit := candidatesNumber + 1
+
+	//Transfer e.votes to a preference map with type "int"
+	for i := 0; i < islandsNumber-1; i++ {
+		for j := 0; j < candidatesNumber-1; j++ {
+			for k := 0; k < candidatesNumber-1; k++ {
```
QFSW: Why are we doing `candidatesNumber-1`? Won't this miss the last candidate?

Ning-72: because index of the first candidate is 0

QFSW: Yes, it is, but you are also doing `<` as your condition which is non exclusive
```go
for i := 0; i < 5; i++
```
Will run 5 times, with `i = {0, 1, 2, 3, 4}`
Don't we want it to run `islandsNumber` times?

QFSW (dismissed):

internal/common/voting/election.go
Ning-72: Oh I'm sorry about this, as I didn't notice that when I modified the loop syntax, I'll modify it immediately.

Darrekt (commented):
There's a lot of loops with single-letter variables, makes it really hard to know what's going on. I'll trust that you've implemented borda count, but could you write a test to show it works?

Ning-72:
> There's a lot of loops with single-letter variables, makes it really hard to know what's going on. I'll trust that you've implemented borda count, but could you write a test to show it works?

Sure I'll do that soon.

QFSW:
@preetl @Neelesh99 Do you mind checking that this would work correctly with the rest of IIGO?

internal/common/voting/election.go
```diff
@@ -55,8 +55,90 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+
+			for j = 0; j < candidatesNumber; j++ {
+				if maxlim > order[j] {
+					maxlim = order[j]
+					index[i] = j
+				}
+			}
+
+			j = index[i]
+			order[j] = 101
```
QFSW: Magic number? Is this maxlim + 1? If so, use that instead as it's clearer :)

internal/common/voting/election.go
```diff
@@ -55,8 +55,90 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+	for k, v := range votesLayoutElect {
+		j := 0
+
+		for t := 0; t < candidatesNumber; t++ {
+			order[t] = v[t]
+		}
+
+		for i := 0; i < candidatesNumber; i++ {
+
+			maxlim := 100
```
QFSW: Try to avoid initializing things with "magic numbers" (so arbitrary constants with no explanation) but they're not always easily avoidable

internal/common/voting/election.go
```diff
@@ -55,8 +55,90 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+	islandsNumber = len(votesSliceSquare)
+	scoreInit := candidatesNumber + 1
+
+	//Transfer e.votes to a preference map with type "int"
+	for i := 0; i < islandsNumber; i++ {
+		for j := 0; j < candidatesNumber; j++ {
+			for k := 0; k < candidatesNumber; k++ {
+				if votesSliceSquare[i][j] == e.islandsToVote[k] {
+					votesLayoutElect[i][k] = scoreInit
+					scoreInit = scoreInit - 1
```
QFSW: `scoreInit -= 1` or even `scoreInit--` would be extra clean

internal/common/voting/election.go
```diff
@@ -55,8 +55,90 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+			score[itrans] = s
+			s++
+		}
+
+		preferenceMap[k] = score
+
+	}
+
+	//Calculate the final score for all candidates and ditermine the winner.
+	var FinalScore []int
```
QFSW: Remember to use `camelCase` for locals

internal/common/voting/election.go
Ning-72: fixed

internal/common/voting/election.go
Ning-72: fixed

internal/common/voting/election.go
Ning-72: set a maxlimitation randomly just to sort the data in preference list from small to large, I'll try to avoid this.

Ning-72:
> There's a lot of loops with single-letter variables, makes it really hard to know what's going on. I'll trust that you've implemented borda count, but could you write a test to show it works?

I tested the code by giving a random input, and below is the test information, I think it works properly. 

Input two-dimensional slice:  [[6 4 5 3 2 1] [1 4 5 3 2 6] [6 4 1 3 2 5] [6 4 5 1 2 3] [3 4 5 6 2 1] [2 4 5 3 6 1]]
Input candidate ID list:  [1 2 3 4 5 6]
Candidate number is  6
Island number is  6
Initialized votesLayoutElect map is map[1:[0 0 0 0 0 0] 2:[0 0 0 0 0 0] 3:[0 0 0 0 0 0] 4:[0 0 0 0 0 0] 5:[0 0 0 0 0 0] 6:[0 0 0 0 0 0]]
Transferred votesLayoutElect map from input is  map[1:[2 3 4 6 5 7] 2:[7 3 4 6 5 2] 3:[5 3 4 6 2 7] 4:[4 3 2 6 5 7] 5:[2 3 7 6 5 4] 6:[2 7 4 6 5 3]]
Sort the votesLayoutElect map as preference map as map[1:[1 2 3 5 4 6] 2:[6 2 3 5 4 1] 3:[4 2 3 5 1 6] 4:[3 2 1 5 4 6] 5:[1 2 6 5 4 3] 6:[1 6 3 5 4 2]]
Calculate the final score of each candidate (in order of candidate1, 2, ...) [16 16 19 30 21 24]
Final winner's clientID is 4

internal/common/voting/election.go
Ning-72: fixed

internal/common/voting/election.go
```diff
@@ -86,8 +86,114 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+				if votesSliceSquare[i][j] == e.islandsToVote[k] {
+					votesLayoutElect[i+1][k] = scoreInit
+					scoreInit -= 1
+				}
+			}
+		}
+	}
+
+	//Sort the preference map in order.
+	var order []int
```
NGovani: can replace with 
`order := make([]int, candidatesNumber`
instead of the for loop

same for the declarations that come after

internal/common/voting/election.go
```diff
@@ -86,8 +86,114 @@ func (e *Election) CloseBallot() shared.ClientID {
 }
 
 func (e *Election) bordaCountResult() shared.ClientID {
-	// TODO implement Borda count winner selection method.
-	return e.pluralityResult()
+	// Implement Borda count winner selection method
+	//(may need to modify if methods design is changed)
+	var candidatesNumber int = 0
+	var islandsNumber int
+	var votesLayoutElect map[int][]int
```
NGovani, line 91: if this represents the islands votes then make it a map[shared.ClientID][]int for standarization

internal/common/voting/election.go
```diff
@@ -86,8 +86,114 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+	votesSliceSquare := e.votes
+	candidatesNumber = len(e.islandsToVote)
+	islandsNumber = len(votesSliceSquare)
+
+	//Initialize votesLayoutMap
+	for i := 1; i < islandsNumber+1; i++ {
+		for j := 0; j < candidatesNumber; j++ {
```
NGovani, line 98: instead of iterating over j, just use 
`votesLayoutElect[i] = make([]int, candiatesNumber)`

preetl (approved):
This should work with IIGO- thanks!

internal/common/voting/election.go
```diff
@@ -86,8 +86,114 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+			votesLayoutElect[i] = append(votesLayoutElect[i], 0)
+		}
+	}
+	//Transfer e.votes to votesLayoutMap with type "int"
+	for i := 0; i < islandsNumber; i++ {
+		scoreInit := candidatesNumber + 1
+		for j := 0; j < candidatesNumber; j++ {
+			for k := 0; k < candidatesNumber; k++ {
+				if votesSliceSquare[i][j] == e.islandsToVote[k] {
```
NGovani, line 107: are there any checks to make sure voteSliceSquare is properly formed? If a client votes has duplicate entries then wouldn't this simply assign the lower score for that candidate.

i.e. if the input was {6,1,2,3,6,4}
the score for candidate 6 wouldn't be  7 it would be 3

internal/common/voting/election.go
```diff
@@ -86,8 +86,114 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+		j := 0
+
+		for t := 0; t < candidatesNumber; t++ {
+			order[t] = v[t]
+		}
+
+		for i := 0; i < candidatesNumber; i++ {
+
+			sum := 0
+			for i := 0; i < candidatesNumber; i++ {
```
NGovani: dont use the same iterator name in an inner for loop
The loop on line 138 also uses i

internal/common/voting/election.go
```diff
@@ -86,8 +86,114 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+				sum = sum + v[i]
+			}
+
+			searcher := sum
+
+			for j = 0; j < candidatesNumber; j++ {
```
NGovani: rename this iterator, j is already declared on line 132

internal/common/voting/election.go
```diff
@@ -86,8 +86,114 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+
+			j = index[i]
+			order[j] = sum
+		}
+
+		itrans := 0
+		s := 1
+		for i := 0; i < candidatesNumber; i++ {
+			itrans = index[i]
+			score[itrans] = s
```
NGovani: instead of using s just use i+1

internal/common/voting/election.go
```diff
@@ -86,8 +86,114 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+		}
+
+		for i := 0; i < candidatesNumber; i++ {
+			preferenceMap[k] = append(preferenceMap[k], score[i])
+		}
+
+	}
+
+	//Calculate the final score for all candidates and ditermine the winner.
+	var finalScore []int
```
NGovani: Again just use the make function

NGovani (changes requested):
there are couple of issue which seem function breaking, specifically the iterator ones fix them and if it still works on your tests then i think it should be good to merge

internal/common/voting/election.go
Ning-72: from my understanding, the input for borda count i.e.  {3,4,5,1,2,6} should be a preference list in which candidates are sorted by popularity, so a candidate should not appear twice in it
if this happens, it should be a problem with GetVoteForElection in baseclient which I think is not completed yet?

internal/common/voting/election.go
Ning-72: thanks!!

internal/common/voting/election.go
Ning-72: it wouldn't be used as output of this function, just convert the votesSliceSquare to map[int][]int for calculation

internal/common/voting/election.go
```diff
@@ -86,8 +86,100 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+	var winnerIndex int
+	winnerIndex = 0
+	for i := 0; i < candidatesNumber; i++ {
+		if maxscore < finalScore[i] {
+			maxscore = finalScore[i]
+			winnerIndex = i
+		}
+	}
+	var winner shared.ClientID
+	type client shared.ClientID
```
NGovani: remove this line

internal/common/voting/election.go
```diff
@@ -86,8 +86,100 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+		}
+	}
+	//Transfer e.votes to votesLayoutMap with type "int"
+	for i := 0; i < islandsNumber; i++ {
+		scoreInit := candidatesNumber + 1
+		for j := 0; j < candidatesNumber; j++ {
+			for k := 0; k < candidatesNumber; k++ {
+				if votesSliceSquare[i][j] == e.islandsToVote[k] {
+					votesLayoutElect[i+1][k] = scoreInit
+					scoreInit -= 1
```
NGovani: change to 
`scoreInit--`

internal/common/voting/election.go
```diff
@@ -86,8 +86,100 @@ func (e *Election) CloseBallot() shared.ClientID {
 }
 
 func (e *Election) bordaCountResult() shared.ClientID {
-	// TODO implement Borda count winner selection method.
-	return e.pluralityResult()
+	// Implement Borda count winner selection method
+	//(may need to modify if methods design is changed)
+	var candidatesNumber int = 0
```
NGovani: remove this line

internal/common/voting/election.go
```diff
@@ -86,8 +86,100 @@ func (e *Election) CloseBallot() shared.ClientID {
...
+	var islandsNumber int
+	var votesLayoutElect map[int][]int
+	votesSliceSquare := e.votes
+	candidatesNumber = len(e.islandsToVote)
```
NGovani: change to
`candidatesNumber := len(e.islandsToVote)`

NGovani (changes requested):
just linter issues which need to be fixed.
Also could you add testing for this.
Look for files in the rest of repo to see examples of testing

NGovani (approved):
This should be fine. 
Considering there's not test i'm going to take your word that it works.

If it breaks, since it is used in IIGO and @preetl has approved it, then both your groups will be responsible for fixing it

Issue #125, opened 2020-12-23
Fix package-level globals in  IIGO


mpardalos:
In [orchestration.go](https://github.com/SOMAS2020/SOMAS2020/blob/main/internal/server/iigointernal/orchestration.go), all state vars are package globals. These should *really* be in `gameState`. 

This was already raised in the review:
![image](https://user-images.githubusercontent.com/2512446/103038248-a9133d80-4576-11eb-9b83-c7aa80c1f7d2.png)

Just mentioning it, but it works, so if no one feels like fixing it, it's fine.

Neelesh99:
PR #138 has a (sort of) fix for this. no one will be hunting for variables every again, its fed directly through signature.

mpardalos:
The globals are still there, but they seem to be basically unused? I'm not sure if that is actually the case. Can we remove them?

Neelesh99:
Not quite unused, they are used to keep persistence of certain variables (e.g. cumulative sanction values) across turns. We could technically move to gamestate, but I think that would be messier than the current implementation where these globals are only used by one function in the file that they are defined in.

lhl2617:
#344 #346 #354 #360

lhl2617:
#363 #362

lhl2617:
resolved!

PR #148, opened 2020-12-27
[IIGO] Move TurnHistory into GameState from global

lhl2617:
# Summary

- `TurnHistory` should not have existed. This should've been in the `GameState`. Partially addresses #125 
- `output.json` now shows correct IIGOHistory instead of the SAME IIGOHistory  for EVERY turn. Each turn has its own individual IIGOHistory now.

## Additional Information

- Found error in IIGO running

## Test Plan

- As there are no test cases that exercise the code I changed, there is no feasible way of testing. It was broken before I fixed it, so I cannot confirm. Feel free to take over this PR and test it.


QFSW:

Neelesh99:
Thanks for doing this @lhl2617, I found the FixMe, in essence, this was us not using error's as intended. We were using them as a generic datatype to pass data rather than their proper use. I've now refactored the return of runIIGO so that it provides both a boolean status and a message (which is logged if that particular run of IIGO was unsuccessful)

Neelesh99 (approved):
I'm happy with these changes, thanks @lhl2617 for doing them. Apologies again or our misuse of error (it was just so convenient ;) ). Feel free to merge.

lhl2617:
@Neelesh99 I'll remove the fixmes then :) thanks!

NGovani (approved):
Yup all the changes make sense and look good

preetl (commented):
posthumous approval! maybe a silly question but i'm just wondering if there will there will be a way of accessing previous turn's IIGOHistory? We will be extending the Judge functionality in our post-MVP tasks to have the option of checking historical events.

lhl2617:
@preetl can you make an issue and assign to me pls

lhl2617:
@preetl `IIGOHistory` is accumulated from start of the game. It is _NOT_ reset at the start of the next turn.

preetl:
> @preetl `IIGOHistory` is accumulated from start of the game. It is _NOT_ reset at the start of the next turn.

@lhl2617 brilliant! that's perfect for our use case :)

PR #146, opened 2020-12-27
[INFRA/CORE] Allow CLI overriding of game config

lhl2617:
# Summary

- Allow overriding of config via CLI instead of just editing config.go file.
- `gameConfig` now passed into `NewSOMASServer` (previously known as SOMASSeverFactory), and stored as a variable.
- Fixed code that depended on global config as well to take in gameConfig from the server.

## Additional Information

Fixed typos here and there, clarified some docs.

## Test Plan

- `go run . --help`
```
Usage of C:\Users\lhlee\Documents\SOMAS\SOMAS2020\SOMAS2020.exe:
  -costOfLiving float
        Subtracted from an islands pool before the next turn.
        This is the simulation-level equivalent to using resources to stay
        alive (e.g. food consumed). These resources are permanently consumed and do
         NOT go into the common pool. Note: this is NOT the same as the tax (default 10)
  -disasterGlobalProb float
        Bernoulli 'p' param. Chance of a disaster occurring. (default 0.1)
  -disasterMagnitudeLambda float
        Exponential rate param for disaster magnitude (default 1)
  -disasterSpatialPDFType int
        Set x,y prob. distribution of the disaster's epicentre (more post MVP)
        0: Uniform

  -disasterXMax float
        Max x bound of archipelago (bounds for possible disaster). (default 10)
  -disasterXMin float
        Min x bound of archipelago (bounds for possible disaster).
  -disasterYMax float
        Max y bound of archipelago (bounds for possible disaster). (default 10)
  -disasterYMin float
        Min y bound of archipelago (bounds for possible disaster).
  -foragingBernoulliProb p
        p param in D variable (see foraging README). Controls prob of catching a deer or not. (default 0.95)
  -foragingDeerGrowthCoefficient float
        Scaling parameter used in the population model. Larger coeff => deer pop. regenerates faster. (default 0.4)
  -foragingExponentialRate lambda
        lambda param in W variable (see foraging README). Controls distribution of deer sizes. (default 1)
  -foragingIncrementalInputDecay float
        Determines decay of incremental input cost of hunting more deer. (default 0.8)
  -foragingMaxDeerPerHunt uint
        Max possible number of deer on a single hunt (regardless of number of participants). (default 4)
  -initialResources float
        The default number of resources at the start of the game. (default 100)
  -maxCriticalConsecutiveTurns uint
        The maximum consecutive turns an island can be in the critical state. (default 3)
  -maxDeerPopulation uint
[...]

PR #150, opened 2020-12-27
[INFRA/CORE] Add necessary Stringer functions for iotas

lhl2617:
# Summary

Add the `String()`, `GoString()`, `MarshalText()`, and `MarshalJSON()` functions for iotas. This is needed to output well-formed strings in JSON & logs.

## Additional Information

- `SpatialPDFType` is left out, this is impl. in #146.
- Not fixing `golint`--these are not caused by my changes.

## Test Plan

Unit tests passed, run succeeded. Logs and output now print actual representative strings instead of bare integers.

### DIFF
--- a/internal/common/voting/election.go
+++ b/internal/common/voting/election.go
@@ -15,6 +15,7 @@ type Election struct {
 	votes         [][]shared.ClientID
 }
 
+// ElectionVotingMethod provides enumerated type for selection of voting system to be used
 type ElectionVotingMethod int
 
 const (
--- a/internal/common/voting/election.go
+++ b/internal/common/voting/election.go
@@ -1,8 +1,11 @@
 package voting
 
 import (
+	"fmt"
+
 	"github.com/SOMAS2020/SOMAS2020/internal/common/baseclient"
 	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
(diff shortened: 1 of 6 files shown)


QFSW:

lhl2617:
CC all infra @Darrekt @QFSW @JamesTev @Eirikalb @Diregie-J @Neelesh99 @777jameskim @P-A-Ha

internal/common/voting/election.go
```diff
@@ -12,14 +15,41 @@ type Election struct {
 	votes         [][]shared.ClientID
 }
 
-type ElectionVotingMethod = int
+type ElectionVotingMethod int
```
Darrekt, line 19: has the = been useless this whole time? or did this actually change something?

Darrekt (approved):
Approving but not merging, will let the other reps have a look at it since it seems important.

Neelesh99 (approved):
LGTM, I've added documentation for the exported fields, hopefully, should resolve linter errors

internal/common/voting/election.go
lhl2617: `type A int` -> A is an integer type
`type B = int` -> B _is_ integer

Very subtle semantic difference--the former means that A is a new type that has the properties of an int. B on the other hand is equivalent to int.

PR #145, opened 2020-12-27
Fix IITO

lhl2617:
# Owned by @Darrekt , originally in #134

# Summary

IITO currently has some issues regarding struct mutation safety and resources being represented as an `int`. This fix standardises it to the `float64` in `shared`.

## Additional Information

Also fixed a bunch of linter errors.

## Test Plan

- Add information on how you tested your changes.

PR #153, opened 2020-12-27
Remove baseclient.clientGameState

mpardalos:
Deprecated in favor of baseclient.serverReadHandle

As requested for #145 by @Darrekt

### DIFF
--- a/internal/common/baseclient/baseclient.go
+++ b/internal/common/baseclient/baseclient.go
@@ -84,7 +84,6 @@ func NewClient(id shared.ClientID) Client {
 // even when new features are added.
 type BaseClient struct {
 	id               shared.ClientID
-	clientGameState  gamestate.ClientGameState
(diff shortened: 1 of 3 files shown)


internal/common/baseclient/baseclient.go
```diff
@@ -126,18 +125,9 @@ func (c *BaseClient) Logf(format string, a ...interface{}) {
 // OPTIONAL. Base should be able to handle it but feel free to implement your own.
 func (c *BaseClient) StartOfTurnUpdate(gameState gamestate.ClientGameState) {
```
Darrekt, line 126: Do we even need `StartOfTurnUpdate` any more?

Darrekt (approved):
Approving and merging. See comment.

internal/common/baseclient/baseclient.go
mpardalos: Yusuf said we should leave it. People might want to log stuff or do some kind of bookkeeping there.

PR #165, opened 2020-12-28
Fix IITO (again)

### LINKED ISSUE #119: Change AcceptGift function in baseclient

Darrekt:
# Summary

The previous implementation of IITO was broken in many ways, which we discovered while trying to fix #119. This PR aims to fix that.

Through a change in the data model, the following is achieved:
1. Clients can no longer mutate the nature of any transaction (a request, offer or response) that is not theirs. 
2. The server now maintains organisation of the transactions, with all sender / receiver information encoded in the data type and key of the respective `map` types.
3. Clients now receive all information for any given round of transactions, so they may make their decisions with all available information instead of on a client-by-client basis.

Server-side sanitisation functions ensure the following:
1. Clients cannot make transactions with dead islands, or themselves.
2. Clients cannot accept more than they were offered. It will be capped to the offer value.
3. Clients cannot offer more than what they have. If they do, the server solves a knapsack packing problem to accept as many offers as it can within the client's total resource capacity.
4. If a client fails to respond to a request or offer, the server fills in an "ignored" reply so that the other client can act on this information.

## Test Plan

The tests are contained in `iito-test.go` in `server`.


internal/server/server_test.go
```diff
@@ -71,29 +70,3 @@ type initTestClient struct {
...
-	}
-
-	createSOMASServer(clientInfos, clientMap)
-
-	for clientID, client := range clientPtrsMap {
-		if !client.initialiseCalled {
-			t.Errorf("Initialise not called for %v", clientID)
-		}
-	}
-}
```
Darrekt: Whoops; accidentally deleted this test case while copy-pasting stuff. will restore.

internal/server/iito.go
```diff
@@ -48,56 +48,103 @@ func (s *SOMASServer) runGiftSession() error {
...
-		giftOfferDict[id], err = client.OfferGifts(giftRequestDict)
+
+// getGiftOffers collects all responses from clients to their requests in a map
+func (s *SOMASServer) getGiftOffers(totalRequests map[shared.ClientID]shared.GiftRequestDict) (map[shared.ClientID]shared.GiftOfferDict, error) {
+	totalOffers := map[shared.ClientID]shared.GiftOfferDict{}
+	for _, thisTeam := range getNonDeadClientIDs(s.gameState.ClientInfos) {
+		// Gather all the requests made to this team
+		requestsToThisTeam := shared.GiftRequestDict{}
+		for fromTeam, indivRequests := range totalRequests {
+			requestsToThisTeam[fromTeam] = indivRequests[thisTeam]
```
NGovani: If  fromTeam does not request anything from thisTeam, then the requestsToThisTeam gets a default initialised GiftRequest which is 0. This is then passed to the clients. So essentially every client gets a request with a value of zero from the teams who requested nothing, which somewhat negates the purpose of the previous sanitisation. 

A similar issue is found on line 128 and on 145

internal/server/iito.go
```diff
@@ -48,56 +48,103 @@ func (s *SOMASServer) runGiftSession() error {
...
-				continue
-			}
-			if receivedByClientDict[idsend] == nil {
-				receivedByClientDict[idsend] = make(shared.GiftDict)
-			}
-			receivedByClientDict[idsend][idto] = giftOffersDict[idto][idsend]
+
+func (s *SOMASServer) sanitiseTeamGiftResponses(responses shared.GiftResponseDict, offers shared.GiftOfferDict) (shared.GiftResponseDict, error) {
+	// Cap each response so that the an island can't accept more than it was offered.
+	for team, response := range responses {
```
NGovani, line 167: this function should also remove responses that don't have a corresponding offer. Potentially a client can simply add a gift response to their returned map.

A simple fix would be to create a clean map in this function and assign to it on line 113 along with the reason. Then return this new map

NGovani (changes requested):
Another TODO is to change the server state once the gift session has been completed, resolving any issues with clients running out of money over multiple offers and then updating the clients with their new resources.

internal/server/iito.go
Darrekt: Nice catch. Didn't know invalid accesses are returned with default values. Fixed with the comma, ok pattern.

NGovani (approved):

Darrekt (commented):
The life and times of ¯\_(ツ)_/¯

Issue #167, opened 2020-12-28
IIGO nil pointer dereference


mpardalos:
I got the clients to contribute something to the common pool to get IIGO to run and this was the result:

<details>
<summary>Output</summary>
<p>

`> go run .  -foragingMaxDeerPerHunt 60 -maxTurns 500 -initialResources 500`

```
2020/12/28 19:53:06 [SERVER]: start runTurn
2020/12/28 19:53:06 [SERVER]: TURN: 1, Season: 1
2020/12/28 19:53:06 [SERVER]: start runOrgs
2020/12/28 19:53:06 [SERVER]: start runIIGO
2020/12/28 19:53:06 [SERVER]: Could not run IIGO since President has no resources to spend
2020/12/28 19:53:06 [SERVER]: finish runIIGO
2020/12/28 19:53:06 [SERVER]: start runIIFO
2020/12/28 19:53:06 [SERVER]: start runPredictionSession
2020/12/28 19:53:06 [Team1]: Final Prediction: [{1.2 1.2 1.2 1 100}]
2020/12/28 19:53:06 [Team2]: Final Prediction: [{1.2 1.2 1.2 1 100}]
2020/12/28 19:53:06 [Team3]: Final Prediction: [{1.2 1.2 1.2 1 100}]
2020/12/28 19:53:06 [Team4]: Final Prediction: [{1.2 1.2 1.2 1 100}]
2020/12/28 19:53:06 [Team5]: Final Prediction: [{1.2 1.2 1.2 1 100}]
2020/12/28 19:53:06 [Team6]: Final Prediction: [{1.2 1.2 1.2 1 100}]
2020/12/28 19:53:06 [SERVER]: finish runPredictionSession
2020/12/28 19:53:06 [SERVER]: Run Forage Predictions
2020/12/28 19:53:06 [SERVER]: Getting Forage Information
2020/12/28 19:53:06 [SERVER]: Distributing Forage Information
2020/12/28 19:53:06 [SERVER]: Finish Running Forage Predictions
2020/12/28 19:53:06 [SERVER]: finish runIIFO
2020/12/28 19:53:06 [SERVER]: start runIITO
2020/12/28 19:53:06 [SERVER]: start runGiftSession
[...]

PR #168, opened 2020-12-28
[INFRA/IIGO] Fixed Nil pointer issue mentioned in issue #167

Neelesh99:
# Summary

Nil pointers were being passed to IIGO from base clients. Base roles are now passed instead.

Some functions signatures in the interface and their implementations were misaligned. They've now been aligned and data is fed appropriately.

## Test Plan

Passes all tests.

### DIFF
--- a/internal/common/baseclient/baseclient_iigo.go
+++ b/internal/common/baseclient/baseclient_iigo.go
@@ -30,17 +30,17 @@ func (c *BaseClient) RuleProposal() string {
 
 // GetClientPresidentPointer is called by IIGO to get the client's implementation of the President Role
 func (c *BaseClient) GetClientPresidentPointer() roles.President {
-	return nil
+	return &BasePresident{}
 }
 
 // GetClientJudgePointer is called by IIGO to get the client's implementation of the Judge Role
 func (c *BaseClient) GetClientJudgePointer() roles.Judge {
-	return nil
+	return &BaseJudge{}
 }
 
 // GetClientSpeakerPointer is called by IIGO to get the client's implementation of the Speaker Role
 func (c *BaseClient) GetClientSpeakerPointer() roles.Speaker {
-	return nil
+	return &BaseSpeaker{}
 }
 
 func (c *BaseClient) TaxTaken(shared.Resources) {
--- a/internal/common/roles/judge.go
+++ b/internal/common/roles/judge.go
@@ -13,8 +13,8 @@ type EvaluationReturn struct {
 	Evaluations []bool
 }
 type Judge interface {
-	PayPresident() (shared.Resources, bool)
+	PayPresident(presidentSalary shared.Resources) (shared.Resources, bool)
 	InspectHistory(iigoHistory []shared.Accountability) (map[shared.ClientID]EvaluationReturn, bool)
-	DeclareSpeakerPerformance() (result bool, didRole bool)
-	DeclarePresidentPerformance() (result bool, didRole bool)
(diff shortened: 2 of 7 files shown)


QFSW:

internal/common/roles/judge.go
```diff
@@ -13,8 +13,8 @@ type EvaluationReturn struct {
 	Evaluations []bool
 }
 type Judge interface {
-	PayPresident() (shared.Resources, bool)
+	PayPresident(presidentSalary shared.Resources) (shared.Resources, bool)
 	InspectHistory(iigoHistory []shared.Accountability) (map[shared.ClientID]EvaluationReturn, bool)
-	DeclareSpeakerPerformance() (result bool, didRole bool)
-	DeclarePresidentPerformance() (result bool, didRole bool)
+	DeclareSpeakerPerformance(inspectBallot bool, conductedRole bool) (result bool, didRole bool)
+	DeclarePresidentPerformance(inspectBallot bool, conductedRole bool) (result bool, didRole bool)
```
mpardalos, line 19: Are these changes related to the nil pointer issue, or is this about something else?

internal/common/baseclient/baseclient_iigo.go
```diff
@@ -30,17 +30,17 @@ func (c *BaseClient) RuleProposal() string {
...
 func (c *BaseClient) GetClientJudgePointer() roles.Judge {
-	return nil
+	return &BaseJudge{}
 }
 
 // GetClientSpeakerPointer is called by IIGO to get the client's implementation of the Speaker Role
 func (c *BaseClient) GetClientSpeakerPointer() roles.Speaker {
-	return nil
+	return &BaseSpeaker{}
 }
```
mpardalos, line 44: Looks like this fixes the issue, however there should be a nil check in IIGO to prevent custom clients from giving a nil pointer to an official, you could either say that implies they're "unfit to rule" i.e. they don't know how to handle the role, or just panic with an explanation so we can figure it out if it happens again.

mpardalos (commented):
Looks like it fixes the issue. However it contains changes that might not be directly related. Maybe mention these changes in the description.

internal/common/roles/judge.go
Neelesh99: Yeah soz about these, we slightly misaligned the function signatures to their implementations this is a fix for that.

internal/common/baseclient/baseclient_iigo.go
Neelesh99: Ok, thats a good plan, implementing now

internal/common/roles/judge.go
mpardalos: That's fine, just edit the PR description to mention this.

internal/server/iigointernal/orchestration.go
```diff
@@ -85,9 +85,9 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
 	presidentPointer = iigoClients[PresidentIDGlobal].GetClientPresidentPointer()
 
 	// Initialise iigointernal with their clientVersions
-	judicialBranch.clientJudge = judgePointer
-	executiveBranch.clientPresident = presidentPointer
-	legislativeBranch.clientSpeaker = speakerPointer
+	judicialBranch.loadClientJudge(judgePointer)
+	executiveBranch.loadClientPresident(presidentPointer)
+	legislativeBranch.loadClientSpeaker(speakerPointer)
```
mpardalos, line 90: This looks good to me.

mpardalos (approved):
image

PR #173, opened 2020-12-29
Fixes around IIGO

mpardalos:
----

# Summary

* `runIIGOAllocations` was never called, added it to iigo end of turn
* `basePresident` would set common pool resource requests to `NaN` if they were all 0

## Test Plan

- Manually tested

### DIFF
--- a/internal/common/baseclient/basepresident.go
+++ b/internal/common/baseclient/basepresident.go
@@ -17,7 +17,7 @@ func (p *BasePresident) EvaluateAllocationRequests(resourceRequest map[shared.Cl
 		requestSum += request
 	}
 
-	if requestSum < 0.75*availCommonPool {
+	if requestSum < 0.75*availCommonPool || requestSum == 0 {
 		resourceAllocation = resourceRequest
 	} else {
 		for id, request := range resourceRequest {
--- a/internal/server/iigo.go
+++ b/internal/server/iigo.go
@@ -29,17 +29,9 @@ func (s *SOMASServer) updateIIGOTurnHistory(clientID shared.ClientID, pairs []ru
 	)
 }
 
-func (s *SOMASServer) runIIGOEndOfTurn() error {
-	s.logf("start runIIGOEndOfTurn")
-	defer s.logf("finish runIIGOEndOfTurn")
-
-	s.runIIGOAllocations()
-	s.runIIGOTax()
-
-	return nil
-}
-
-func (s *SOMASServer) runIIGOTax() {
+func (s *SOMASServer) runIIGOTax() error {
+	s.logf("start runIIGOTax")
+	defer s.logf("finish runIIGOTax")
 	clientMap := getNonDeadClients(s.gameState.ClientInfos, s.clientMap)
 	for clientID, v := range clientMap {
 		tax := v.GetTaxContribution()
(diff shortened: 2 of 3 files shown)


QFSW:

Neelesh99 (changes requested):
Hi, thanks for this, but the runIIGOAllocations() needs to be run separately from runIIGOTax(), in our infinite wisdom we decided to create the fully separate function but forgot to call it. @preetl , @Spruudis, could you confirm the sequencing for this in Turn.go and we can implement it.

preetl:
As I understand it, all deductions from islands' personal pools should happen before any incomings, so that islands make their choices on how much to gift/forage/tax with the resources they already had, not 'new resources' that they just got for example from deer hunting. With this in mind, IIGO taxes should be put into the common pool first, and then gifting and foraging (not sure which is first out of these two - I believe it is undecided). After this, all incomings to the islands' personal pools can happen i.e. taking from the common pool, receiving gifts and receiving foraging returns (once again, order of these 3 is undecided but perhaps inconsequential and easy to change). I don't think it matters when IIFO actions (predictions) happen.

Essentially, we need to keep end of turn actions flexible enough to be able to play with the ordering so we should have separate funcs for each outgoing and incoming to personal pools to be able to do this. Each of these should be called in endOfTurn() in turn.go rather than calling a runOrgsEndOfTurn() function.

internal/common/baseclient/basepresident.go
```diff
@@ -17,7 +17,7 @@ func (p *BasePresident) EvaluateAllocationRequests(resourceRequest map[shared.Cl
 		requestSum += request
 	}
 
-	if requestSum < 0.75*availCommonPool {
+	if requestSum < 0.75*availCommonPool || requestSum == 0 {
```
preetl, line 20: Good catch! Thanks!

mpardalos:
The way I understand it, the turn happens like this:

1. StartOfTurn()
2. IIGO
3. IITO
4. IIFO
5. IIGOEndOfTurn
6. IIFOEndOfTurn
7. IITOEndOfTurn
8. forage
9. disaster
10. cost of living

See the following for how this order is found:
* https://github.com/SOMAS2020/SOMAS2020/blob/c01785a0b69178fb44edd1a0ca144ca19b68702b/internal/server/turn.go#L9
* https://github.com/SOMAS2020/SOMAS2020/blob/c01785a0b69178fb44edd1a0ca144ca19b68702b/internal/server/turn.go#L30
* https://github.com/SOMAS2020/SOMAS2020/blob/c01785a0b69178fb44edd1a0ca144ca19b68702b/internal/server/turn.go#L37
* https://github.com/SOMAS2020/SOMAS2020/blob/c01785a0b69178fb44edd1a0ca144ca19b68702b/internal/server/turn.go#L57
* https://github.com/SOMAS2020/SOMAS2020/blob/c01785a0b69178fb44edd1a0ca144ca19b68702b/internal/server/turn.go#L93

The turn is currently split across the above functions. I think it would be easier to just call all of the org/foraging/etc. functions directly from `SOMASServer.Turn` so that the order is obvious. @preetl, I'm sure that the current order is wrong in some way cause I don't think it matches what you described. 

Should I make this change? I.e. get rid of the various functions called in `SOMASServer.Turn` and just call the org functions directly there. We can then make it be exactly what the design mandates.

Personally I am not really familiar with how it's supposed to work, I was just trying to do the agent and realised the method had not been called. I'd appreciate it if someone with more experience explains how this is supposed to work.

CC: @lhl2617

NGovani (approved):
Looks good, the more modular the better.

Neelesh99 (approved):
Looks good, thanks for this ;)

PR #175, opened 2020-12-29
Reformatted End Of Turn Events

preetl:
# Summary

In response to PR #173. A more flexible end of turn execution that lets us adjust ordering if necessary. I believe all personal pool outgoings should happen before personal pool incomings and this setup will allow that.

### DIFF
--- a/internal/server/turn.go
+++ b/internal/server/turn.go
@@ -54,18 +54,31 @@ func (s *SOMASServer) runOrgs() error {
 }
 
 // endOfTurn performs end of turn updates
+// TODO: organise order of end of turn actions
 func (s *SOMASServer) endOfTurn() error {
 	s.logf("start endOfTurn")
 	defer s.logf("finish endOfTurn")
 
-	err := s.runOrgsEndOfTurn()
-	if err != nil {
-		return errors.Errorf("Failed to run orgs end of turn: %v", err)
+	if err := s.runIIGOTax(); err != nil {
+		return errors.Errorf("Failed to put taxes into common pool at end of turn: %v", err)
 	}
 
-	err = s.runForage()
-	if err != nil {
+	if err := s.runIIFOEndOfTurn(); err != nil {
+		return errors.Errorf("IIFO EndOfTurn error: %v", err)
+	}
+
+	// TODO: break IITO down into giving gifts and receiving gifts
+	if err := s.runIITOEndOfTurn(); err != nil {
+		return errors.Errorf("IITO EndOfTurn error: %v", err)
+	}
+
+	// TODO : break foraging down into foraging investments and foraging returns
+	if err := s.runForage(); err != nil {
 		return errors.Errorf("Failed to run hunt at end of turn: %v", err)
+	}
+
+	if err := s.runIIGOAllocations(); err != nil {
+		return errors.Errorf("Failed to get common pool allocations at end of turn: %v", err)
 	}
 
 	// probe for disaster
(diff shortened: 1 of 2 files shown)


internal/server/turn.go
```diff
@@ -54,20 +54,33 @@ func (s *SOMASServer) runOrgs() error {
...
-	err := s.runOrgsEndOfTurn()
-	if err != nil {
-		return errors.Errorf("Failed to run orgs end of turn: %v", err)
+	if err := s.runIIGOTax(); err != nil {
+		return errors.Errorf("Failed to put taxes into common pool at end of turn: %v", err)
 	}
 
-	err = s.runForage()
-	if err != nil {
+	if err := s.runIIFOEndOfTurn(); err != nil {
```
NGovani, line 66: IIFO and IITO do not currently have any end of turn action currently and I doubt they will in the future.
Would you mind deprecating these two functions, just delete them from here and there declaration in the iito and iifo files

NGovani (changes requested):
Everything looks good, one minor change

internal/server/turn.go
preetl: Sure, I can do that! Just to check then, are gifts actually given and received during the IITO session? I thought it was just like an agreement that was made but it actually happened at the end of the turn.

NGovani: With actions being deprecated it makes more sense to just give the clients the resources as soon as they make the agreement, I don't see any reason to wait till the end of turn.

Currently IITO doesn't update the client state but there is a PR with an overhaul that will include that soon.

preetl: I think maybe we should check with the rest of design on this as I recall there being conversations about how it should appear that islands make decisions on what to do with the resources that they have concurrently. The purpose of this was to ensure that a decision on one action wasn't influenced by the outcome of another. This can be achieved by doing all actions that subtract from their pool before actions that add to their pool. 

Just to say, not sure that this reasoning is correct anymore but I remember this being decided on.

preetl: I'll remove IIFO end of turn actions in the meantime
Edit: looks like there are functions called in IIFOendofturn() so i won't delete it.

NGovani: Mm that does seem like a fair point. Alright we can just merge these changes then a revisit it later. Is that okay?

preetl: sure! it's just a pr into @mpardalos ' branch

internal/server/turn.go
```diff
@@ -54,20 +54,33 @@ func (s *SOMASServer) runOrgs() error {
...
+// TODO: organise order of end of turn actions
 func (s *SOMASServer) endOfTurn() error {
 	s.logf("start endOfTurn")
 	defer s.logf("finish endOfTurn")
 
-	err := s.runOrgsEndOfTurn()
-	if err != nil {
-		return errors.Errorf("Failed to run orgs end of turn: %v", err)
+	if err := s.runIIGOTax(); err != nil {
+		return errors.Errorf("Failed to put taxes into common pool at end of turn: %v", err)
```
mpardalos, line 63: Is there a reason that IIGO is split in 2 sections?

internal/server/turn.go
```diff
@@ -54,20 +54,33 @@ func (s *SOMASServer) runOrgs() error {
...
+
+	// TODO : break foraging down into foraging investments and foraging returns
+	if err := s.runForage(); err != nil {
 		return errors.Errorf("Failed to run hunt at end of turn: %v", err)
 	}
 
+	if err := s.runIIGOAllocations(); err != nil {
+		return errors.Errorf("Failed to get common pool allocations at end of turn: %v", err)
+	}
+
```
mpardalos, line 83: Is foraging meant to run before allocations?

internal/server/turn.go
```diff
@@ -89,26 +102,6 @@ func (s *SOMASServer) endOfTurn() error {
...
-		return errors.Errorf("IIFO EndOfTurn error: %v", err)
-	}
-
-	if err := s.runIITOEndOfTurn(); err != nil {
-		return errors.Errorf("IITO EndOfTurn error: %v", err)
-	}
-
-	return nil
-}
-
```
mpardalos, line 111: Completely agree with getting rid of this.

NGovani (approved):

internal/server/turn.go
```diff
@@ -54,20 +54,33 @@ func (s *SOMASServer) runOrgs() error {
 }
 
 // endOfTurn performs end of turn updates
+// TODO: organise order of end of turn actions
 func (s *SOMASServer) endOfTurn() error {
```
mpardalos, line 58: Should we maybe inline this function into `runTurn`? We could also inline `runOrgs` there. 

I don't think the distinction between runOrgs/endOfTurn is really useful.

mpardalos (dismissed):
Looks good implementation-wise, I can't comment on if the order of events is correct because I'm not involved in design. 

Made some suggestions for a couple of further changes.

PR #189, opened 2020-12-29
Adding initial common pool configuration

Spruudis:
Added option to configure the common-pool value at the start of the game


QFSW:

lhl2617 (approved):
image

would be nice to see test plan with example runs with different args for this config

PR #190, opened 2020-12-29
[VIS/WASM] Fix bug in which args reset do not take effect

lhl2617:
# Summary

As title. This is because we only pass in non-default arguments; when the Go program has replaced the defaults with the new arguments it no longer resets the flag to its default.

It is possible to do the reset in the Go program, but it is more tedious.

This PR passes in _all_ flags and values, default or not.

## Test Plan

No longer possible to repro bug.


QFSW:

preetl (approved):
bye bye bug

PR #193, opened 2020-12-30
New System Of Power Transfer In IIGO

preetl:
# Summary

This is a change to how new roles of power in IIGO are elected. For MVP, an election was called at the end of every turn but islands now have the power to decide whether to initiate a power transfer for the role they are responsible for i.e. the Speaker decides when to call an election for a new Judge, the Judge for the President, and the President for the Speaker.

(Part 1 of the accountability cycle)

## Additional Information

Moved the ElectionVotingMethod type and ElectionSettings struct to shared to accommodate this change. Also, corresponding BaseClient functions have been added.

## Test Plan

Tests to come!


QFSW:

internal/server/iigointernal/executive.go
```diff
@@ -12,15 +12,16 @@ import (
...
+	clientPresident     roles.President
+	budget              shared.Resources
+	speakerSalary       shared.Resources
+	RulesProposals      []string
+	ResourceRequests    map[shared.ClientID]shared.Resources
+	speakerTurnsInPower int
 }
 
-// loadClientJudge checks client pointer is good and if not panics
+// loadClientPresident checks client pointer is good and if not panics
```
Neelesh99, line 24: Ty for this

Neelesh99 (approved):
Looks good to me

PR #188, opened 2020-12-29
[IIGO/Judge] Post-MVP updates for Judge 

Neelesh99:
# Summary

Judge is now able to:

- Sanction islands for breaking rules
> - Set Sanction tier thresholds
> - Set sanction score penalised for breaking any particular rule

- Pardon islands who are currently sanctioned
> - Select exactly which island and what sanction to pardon
> - Costs resources ?

- Retro-actively punish islands who's actions escaped evaluation by a previous judge
> - Judiciary branch now keeps track of what actions have gone unchecked from (currently 3) a configurable number of rounds ago
> - Islands can choose to evaluate these unchecked actions with the same freedom to ignore any they want

## Additional Information

This implementation is pending an Accountability cycle expansion, so it may look quite different by the end of post-MVP changes, I'd recommend any teams hold off on implementing their judges based on this code until the accountability and new orchestration is in place.

## Test Plan

- Lots of tests added ;)

P.S. If anyone from Team1 can help me move the sanctionCacheDepth and historyCacheDepth (top of judiciary.go) to the params/config system you guys have made that would be awesome. Might want to wait till all of post-MVP IIGO is in place though.

P.S.P.S. Sorry for the size of this PR again.

PR #196, opened 2020-12-30
[INFRA/IIGO] Sanction threshold setting and rule violation sanction scoring added [1/4]

Neelesh99:
# Summary

As per comments in #188, breaking up PR.

This PR allows Judge role to:

- Set Sanction tier thresholds
- Set sanction score penalised for breaking any particular rule

## Additional Information

Part 1 of 4 (probably, might be 3)
The other parts will build on this so we need to make sure this one is merged first or we risk some merge conflicts.

## Test Plan

Added unit tests for more complex function. More to come.


QFSW:

preetl (approved):
Lovely! The linter errors seem to be because of stuff that you've had to take out to break down the PR so no problem :)

NGovani:
Thank you so much for splitting these!

internal/common/baseclient/basejudge.go
```diff
@@ -9,6 +9,18 @@ import (
 type BaseJudge struct {
 }
 
+// GetRuleViolationSeverity returns a custom map of named rules and how severe the sanction should be for transgressing them
+// If a rule is not named here, the default sanction value added is 1
+func (j *BaseJudge) GetRuleViolationSeverity() map[string]roles.IIGOSanctionScore {
+	return map[string]roles.IIGOSanctionScore{}
```
NGovani, line 15: were rules not changed to be an enum. 

After reading the rest i assume you expect the client to call .string on the rule.
Would it be possible to get an example of how a client would interact with these features. Potentially in a future PR after these 4 are complete

internal/common/roles/judge.go
```diff
@@ -8,13 +8,41 @@ import (
...
+type IIGOSanctionTier int
+
+// Provides enumerated tiers for IIGO sanctions
+const (
+	SanctionTier1 IIGOSanctionTier = iota
+	SanctionTier2
+	SanctionTier3
+	SanctionTier4
+	SanctionTier5
+	None
```
lhl2617: `None` will be a global const in this package. Let's name it something more specific to the type

internal/server/iigointernal/judiciary_test.go
```diff
@@ -0,0 +1,309 @@
+package iigointernal
+
+import (
+	"fmt"
```
lhl2617, line 4: can you run `goimport`?

internal/common/baseclient/basejudge.go
Neelesh99: Right now whether rules are enum's is an open question since design haven't yet closed the door to invented rules. Will update if that changes.

Neelesh99: Yes I'll include examples of use in a future PR. In fact, could we have like a "Demo" client set up for the infra teams to demo their features (rather than in base client which we'd prefer to keep plain and stable?) @NGovani @lhl2617

internal/server/iigointernal/judiciary_test.go
Neelesh99: Not sure how to run the command from goland, I've fixed the imports so that the linter is less upset

internal/common/roles/judge.go
Neelesh99: Agreed and changed

internal/server/iigointernal/judiciary_test.go
```diff
@@ -0,0 +1,310 @@
...
+			if !reflect.DeepEqual(res, tc.expectedVal) {
+				t.Errorf("Expected final transgressions to be %v got %v", tc.expectedVal, res)
+			}
+		})
+	}
+}
+
+// Utility functions for testing //
+
+func generateRuleStore() map[string]rules.RuleMatrix {
```
lhl2617, line 73: remove if unused?

NGovani: I believe those are gonna be used in the upcoming PRs, considering the situation I think we can let it go this time

NGovani (approved):
Looks good we can move onto the next PR

PR #197, opened 2020-12-30
[INFRA/IIGO] Pardoning tooling added with judge methods [2/4]

Neelesh99:
# Summary

As per comments in #188 breaking up PR.

This PR gives the judiciary branch the tools to:
- Select exactly which island and what sanction to pardon

## Additional Information

Part 2 of 4 
The other parts will build on this so we need to make sure this one is merged seconds or we risk some merge conflicts.

## Test Plan

Unit tests added.


QFSW:

Neelesh99:
Evidently, that did not fix that test

internal/server/iigointernal/judiciary.go
```diff
@@ -99,3 +115,125 @@ func (j *judiciary) appointNextPresident(currentPresident shared.ClientID, allIs
...
+		if clientVal, ok := clientSanctionMap[k]; ok {
+			defaultMap[k] = clientVal
+		}
+	}
+	return defaultMap
+}
+
+// getIslandSanctionTier if statement based evaluator for which sanction tier a particular score is in
+func getIslandSanctionTier(islandScore roles.IIGOSanctionScore, scoreMap map[roles.IIGOSanctionTier]roles.IIGOSanctionScore) roles.IIGOSanctionTier {
+	if islandScore <= scoreMap[roles.SanctionTier1] {
```
lhl2617, line 157: ok here. feel free to use `https://godoc.org/github.com/zenixls2/stl` especially lowerbound if you need more cases

Neelesh99:
Ok this new function implementPardons does actually fix whatever deficiencies there were in the old checkPardons functions. Tests have been adapted and all pass consistently now. Thanks @lhl2617 for your help ;).

NGovani:
@lhl2617 if we merge with main, will it remove the changes from the first PR in this PR

Neelesh99:
Oh crap, I thought it would only pick up the differences from main because they share the same commits?

NGovani:
I think it will once we update your branch, thats what i meant. Not to merge these changes into main.
Ye I get how my first comment may have been confusing, sorry

Neelesh99:
Well there's only one way to find out ;)

internal/common/baseclient/basejudge.go
```diff
@@ -64,6 +64,10 @@ func (j *BaseJudge) InspectHistory(iigoHistory []shared.Accountability) (map[sha
 	return outputMap, true
 }
 
+func (j *BaseJudge) GetPardonedIslands(currentSanctions map[int][]roles.Sanction) map[int][]bool {
```
NGovani, line 67: clientID and not int?
also a comment woould be great as to the usage of this function

internal/server/iigointernal/judiciary.go
```diff
@@ -109,6 +116,18 @@ func (j *judiciary) appointNextPresident(currentPresident shared.ClientID, allIs
 	return nextPresident
 }
 
+// cycleSanctionCache rolls the sanction cahce one turn forward (effectively dropping any sanctions longer than the depth)
+func (j *judiciary) cycleSanctionCache() {
+	newMap := j.localSanctionCache
+	delete(newMap, sanctionCacheDepth-1)
+	newMapCache := newMap
```
NGovani: this assigns a reference to newMap and thus in the for loop below you're essentially working on the same data.
To copy a map you have to use a for loop. Potentially you could create an empty map and assign to it instead

internal/server/iigointernal/judiciary.go
```diff
@@ -132,3 +151,140 @@ func softMergeSanctionThresholds(clientSanctionMap map[roles.IIGOSanctionTier]ro
 	}
 	return defaultMap
 }
+
+// getIslandSanctionTier if statement based evaluator for which sanction tier a particular score is in
+func getIslandSanctionTier(islandScore roles.IIGOSanctionScore, scoreMap map[roles.IIGOSanctionTier]roles.IIGOSanctionScore) roles.IIGOSanctionTier {
```
NGovani, line 178: this assumes that the sanction costs scale linearly. if sanction tier 4 is less than 3 this would not function as expected. However i assume you dont want that case to ever happen, so it might be worth adding a check for this when the judge returns their scoreMap

internal/server/iigointernal/judiciary.go
```diff
@@ -132,3 +151,140 @@ func softMergeSanctionThresholds(clientSanctionMap map[roles.IIGOSanctionTier]ro
...
+	if islandScore <= scoreMap[roles.SanctionTier1] {
+		return roles.NoSanction
+	} else if islandScore <= scoreMap[roles.SanctionTier2] {
```
NGovani: Shouldn't this be `<` not `<=`. Id assume if someone gets 5 points they should get a sanction of tier 1

internal/server/iigointernal/judiciary.go
```diff
@@ -132,3 +151,140 @@ func softMergeSanctionThresholds(clientSanctionMap map[roles.IIGOSanctionTier]ro
...
+	for i := 0; i < depth; i++ {
+		returnMap[i] = []roles.Sanction{}
+	}
+	return returnMap
+}
+
+func implementPardons(sanctionCache map[int][]roles.Sanction, pardons map[int][]bool, allTeamIds []shared.ClientID) (bool, map[int][]roles.Sanction, map[shared.ClientID][]map[shared.CommunicationFieldName]shared.CommunicationContent) {
+	if validatePardons(sanctionCache, pardons) {
+		finalSanctionCache := sanctionCache
+		communicationsAboutPardons := generateEmptyCommunicationsMap(allTeamIds)
```
NGovani, line 217: is allTeamIds a global store somewhere?

internal/server/iigointernal/judiciary.go
```diff
@@ -132,3 +151,140 @@ func softMergeSanctionThresholds(clientSanctionMap map[roles.IIGOSanctionTier]ro
...
+	returnMap := map[int][]roles.Sanction{}
+	for i := 0; i < depth; i++ {
+		returnMap[i] = []roles.Sanction{}
+	}
+	return returnMap
+}
+
+func implementPardons(sanctionCache map[int][]roles.Sanction, pardons map[int][]bool, allTeamIds []shared.ClientID) (bool, map[int][]roles.Sanction, map[shared.ClientID][]map[shared.CommunicationFieldName]shared.CommunicationContent) {
+	if validatePardons(sanctionCache, pardons) {
+		finalSanctionCache := sanctionCache
```
NGovani, line 216: once again since this is simply a reference meaning  we're both iterating over and changing sanctionCache in the loop below there may be some unintended behaviour

Also would it not be easier for Sanctions to be a map of clientIDs to sanctions tiers and turns left. Unless of course a client can get multiple sanctions.

NGovani (commented):
Couple of things you might want to look at. Looks good I'm trusting since you've done testing that it works as you want.

I'm not approving since I wanna take another look at it, and im going to assume @LH or someone else is gonna find something.

internal/common/baseclient/basejudge.go
Neelesh99: The int's mentioned here in the map are representative of how long ago the sanction was levied. Admittedly an array would've done the trick but the map seemed convenient at the time.

internal/server/iigointernal/judiciary.go
Neelesh99: Sure, I might end up adding a check when the clients give us the map in the first place.

internal/server/iigointernal/judiciary.go
Neelesh99: Hmmm good point, could this be in another pr as well (it s a bit of risky business rebasing the following commits)

internal/server/iigointernal/judiciary.go
Neelesh99: I don't think so, its fed into the function using the shared.TeamIds in the calling function.

internal/server/iigointernal/judiciary.go
Neelesh99: In this case I've rewritten this function with that in mind, so everything is an assignment now hopefully. As before this map is indexed by timestep, not clientID.

internal/server/iigointernal/judiciary.go
Neelesh99: Ah crap this is something that needs an immediate fix, ok I'll patch this along with the others and have fun with rebasing again

internal/server/iigointernal/judiciary.go
NGovani: oh my bad completely missed. Apolgies

internal/server/iigointernal/judiciary.go
NGovani: Ah the second part of my comment was more to do with the type of sanction. Its a struct of ClientID and then some sanction data. Would it not be easier to type it as a map and then not need a list i.e []roles.Sanctions -> roles.Sanctions(which is now a map)

internal/server/iigointernal/judiciary.go
Neelesh99: Done

internal/server/iigointernal/judiciary.go
Neelesh99: It's fine rebasing seems unavoidable

internal/server/iigointernal/judiciary.go
Neelesh99: Hmmmm that's a good thought. Could we do that in another PR, this rebasing thing hasn't gone well.

NGovani: Sure, as long we have a list of all changes that need to be made. This one is also a lower priority.

NGovani (approved):
In the interest of time I'm going to approve this merge, though it does seem fine.

Could you go through and fix all the annotation errors in the PR, and then we can merge and move on the next one

Neelesh99:
> In the interest of time I'm going to approve this merge, though it does seem fine.
> 
> Could you go through and fix all the annotation errors in the PR, and then we can merge and move on the next one

I think all the lines you've annotated have been addressed or at least made note of for the future.

PR #198, opened 2020-12-30
[INFRA/IIGO] Tooling for historical retribution added [3/4]

Neelesh99:
# Summary

As per comments in #188, breaking up PR.
This PR gives the judiciary branch the tools to
- Judiciary branch now keeps track of what actions have gone unchecked from (currently 3) a configurable number of rounds ago

## Additional Information

Part 3 of 4
We need to make sure this is merged third or we'll have merge conflicts.

## Test Plan

Added unit testing.


QFSW:

internal/server/iigointernal/judiciary.go
```diff
@@ -128,6 +130,23 @@ func (j *judiciary) cycleSanctionCache() {
 	j.localSanctionCache = newMapReturn
 }
 
+// cycleHistoryCache rolls the history cache (for retributive justice) forward
```
NGovani, line 133: is this function used in the next PR, currently it's unused

internal/server/iigointernal/judiciary.go
```diff
@@ -310,3 +338,75 @@ func removeSanctions(slice []roles.Sanction, s int) []roles.Sanction {
...
+	for i, v := range array {
+		if v == val {
+			return i, true
+		}
+	}
+	return -1, false
+}
+
+// cullCheckedRules removes any entries in the history that have been evaluated in evalResults (for historical retribution)
+func cullCheckedRules(iigoHistory []shared.Accountability, evalResults map[shared.ClientID]roles.EvaluationReturn, rulesCache map[string]rules.RuleMatrix, variableCache map[rules.VariableFieldName]rules.VariableValuePair) []shared.Accountability {
```
NGovani, line 369: So while do think this function works as you want, it would be fantastic to under why do these things. Currently I'm really lost as to why these functions would be of use to a judge. Is there going to be an accompanying document tell teams how the system works or is there no need to understand why this happens.

internal/server/iigointernal/judiciary.go
```diff
@@ -310,3 +338,75 @@ func removeSanctions(slice []roles.Sanction, s int) []roles.Sanction {
 func getDifferenceInLength(slice1 []roles.Sanction, slice2 []roles.Sanction) int {
 	return len(slice1) - len(slice2)
 }
+
+// pickUpRulesByVariable returns a list of rule_id's which are affected by certain variables.
+func pickUpRulesByVariable(variableName rules.VariableFieldName, ruleStore map[string]rules.RuleMatrix, variableMap map[rules.VariableFieldName]rules.VariableValuePair) ([]string, bool) {
```
NGovani, line 343: This function is found int rules package with a different signature. Is there a reason why you cant change the other version and use that instead. It seems like in this format there's going to be confusion as to which function to use at times

NGovani (changes requested):
Firstly Historical Retribution is a fantastic name.

Secondly the PR looks good i assume the 4th one will connect alot of the dots together, just one issue with pickUpRulesByVariable() function being in multiple places

internal/server/iigointernal/judiciary.go
Neelesh99: Yeah I think so, PR4 should deffo use it

internal/server/iigointernal/judiciary.go
Neelesh99: Well I don't mind writing something that explains that function, it won't be exposed to the judge at all, it's part of IIGO's internal auditing system. But I admit it is a little complicated, but kinda needs to be for what it does.

internal/server/iigointernal/judiciary.go
Neelesh99: @preetl Stole it from me and then pushed first XD. No you're right, I should use the rules version. Could I open this up as an issue for after this PR though, just because the whole rebasing goes haywire every time?

NGovani (approved):

PR #203, opened 2020-12-30
Allow clients to control how much to give in GiftTransactions

E-Mac27:
# Summary
Clients previously was forced to give the amount they promised initially. This isn't quite inline with IITO design. Since giftsTransactions execute at the end of turn, clients state have changed.
  
Now they are able to specify if they want to fulfil the amount they initially promised or change it. 

## Test Plan
Oops.


QFSW:

NGovani (approved):

Eirikalb (approved):
LGTM, could you also remove the sendGift and receiveGift methods as these are obsolete as an effect of this?

internal/server/iito.go
```diff
@@ -230,14 +230,19 @@ func (s *SOMASServer) distributeGiftHistory(totalResponses map[shared.ClientID]s
 func (s *SOMASServer) executeTransactions(transactions map[shared.ClientID]shared.GiftResponseDict) {
```
lhl2617, line 230: whomst'd've todo is this

NGovani:
> LGTM, could you also remove the sendGift and receiveGift methods as these are obsolete as an effect of this?

They are still needed to notify the clients that their resources have changed. I believe they were changed in a previous PR to sentGift and recievedGift, just a change in tense

internal/server/iito.go
E-Mac27: errrrrrrrrrrrrrrrrrrrrrrrrrr

PR #205, opened 2020-12-30
[INFRA/IIGO] Monitoring System Implemented For Accountability Cycle

preetl:
# Summary

Implemented the monitor object which is used by the President, Speaker, and Judge to monitor one another's performance. The Speaker monitors the President, the President monitors the Judge, and the Judge monitors the Speaker. This involves checking rules pertaining to the operation of that role.

This draft PR only involves monitoring on the Speaker role but will be extended to the other roles. Extending it mainly involves writing rules so I thought this was a good time to share.

## Additional Information

Part 2 of Accountability Cycle

## Test Plan

Tests to come!
Edit: Some tests included but more to come once this feature is fully implemented!


QFSW:

internal/server/iigointernal/utilities.go
```diff
@@ -44,3 +44,11 @@ func WithdrawFromCommonPool(value shared.Resources, gameState *gamestate.GameSta
 func depositIntoCommonPool(value shared.Resources, state *gamestate.GameState) {
 	state.CommonPool += value
 }
+
```
preetl, line 47: Any suggestions on a less ugly function would be appreciated! I couldn't think of a way around this...

internal/server/iigointernal/utilities.go
```diff
@@ -44,3 +44,11 @@ func WithdrawFromCommonPool(value shared.Resources, gameState *gamestate.GameSta
...
+func boolToFloat(input bool) float64 {
+	if input {
+		return 1
+	} else {
+		return 0
+	}
+}
```
preetl: Any suggestions on a less ugly function would be appreciated! I couldn't think of a way around this...

internal/server/iigointernal/monitoring.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (m *monitor) findRoleToMonitor(roleAccountable baseclient.Client) (shared.ClientID, baseclient.Role) {
+	switch roleAccountable.GetID() {
+	case m.speakerID:
+		return m.presidentID, baseclient.President
+	case m.presidentID:
+		return m.judgeID, baseclient.Judge
+	default:
+		return m.speakerID, baseclient.Speaker
+	}
```
E-Mac27, line 71: Is there a possibility that non-speaker/president/judge can have access to this function?

preetl: Good point. No, it's not exported so it is only available within iigointernal (a serverside package) and it will only be called in orchestration.go using the baseclient.Client pointer corresponding to the assigned speaker, judge and president roles.

E-Mac27 (commented):
Any tests if these works as intended? 

Edit: Ignore this comment. I noticed the Summary. Happy to approve this PR when its ready

internal/server/iigointernal/utilities.go
777jameskim: the else is redundant, but other than that 🤷🏻‍♂️

preetl:
No rush @E-Mac27 but this is ready for a review whenever you are free! Since you last saw this I've added in additional rules and changed some functions to have bool return types so that these rules can be evaluated. I will do a separate PR with tests soon!

internal/common/baseclient/baseclient.go
```diff
@@ -54,6 +54,13 @@ type Client interface {
 	GetGiftResponses(receivedOffers shared.GiftOfferDict) shared.GiftResponseDict
 	UpdateGiftInfo(receivedResponses shared.GiftResponseDict)
 
+<<<<<<< HEAD
+	//IIGO
+	MonitorIIGORole(Role) bool
+
+	//TODO: THESE ARE NOT DONE yet, how do people think we should implement the actual transfer?
+=======
+>>>>>>> main
```
preetl:
```suggestion
	//IIGO
	MonitorIIGORole(Role) bool

	//TODO: THESE ARE NOT DONE yet, how do people think we should implement the actual transfer?
```

internal/server/iigointernal/utilities.go
preetl: ah yes! i've removed it now :)

internal/common/baseclient/baseclient.go
```diff
@@ -149,3 +156,8 @@ func (c *BaseClient) ReceiveCommunication(sender shared.ClientID, data map[share
 func (c *BaseClient) GetCommunications() *map[shared.ClientID][]map[shared.CommunicationFieldName]shared.CommunicationContent {
 	return &c.Communications
 }
+
+//MonitorIIGORole decides whether to perform monitoring on the
+func (c *BaseClient) MonitorIIGORole(roleName Role) bool {
```
E-Mac27: Please add "COMPULSORY" or "OPTIONAL"

E-Mac27 (approved):
I'd prefer if you add tests in the same PR since code will be out of place when reviewing or people will forget. But that aside: please fix the comment about Compulsory / Optional.

internal/common/baseclient/baseclient.go
preetl: done!

preetl:
@E-Mac27 That makes complete sense. I've added a test file with about 5 tests. I have a few extensions to this implementation so there will be more on the next PR!

PR #208, opened 2020-12-30
[IIGO] Added role function returns, reworked role salaries

akowal3:
# Summary

Added:

- president return types
- speaker return types
- president islands alive safety improvement
- reworked salaries and budgets for roles
- added variable action costs to all roles
- moved salaries into islands' private pools
hopefully, I didn't miss anything...

## Auxilary info
Sorry for a large PR...

## Test Plan

- Passes all the tests - old (adapted) and new


QFSW:

NGovani:
Is it possible to split this PR up into smaller parts. Reviewing 1.2k takes a long time.
There's a guide here: https://www.thedroidsonroids.com/blog/splitting-pull-request
Also @Neelesh99 had to do this before so he may be able to help if you ask nicely.

Spruudis:
> Is it possible to split this PR up into smaller parts. Reviewing 1.2k takes a long time.
> There's a guide here: https://www.thedroidsonroids.com/blog/splitting-pull-request
> Also @Neelesh99 had to do this before so he may be able to help if you ask nicely.

The reason why this PR is so big is that most of it changes one thing in every function call. Budget affects almost every part of the system. (Also 600 lines of those are tests).

internal/server/iigointernal/executive.go
```diff
@@ -48,73 +48,118 @@ func (e *executive) setAllocationRequest(resourceRequests map[shared.ClientID]sh
...
 // Send Tax map all the remaining islands
 // Called by orchestration at the end of the turn
-func (e *executive) getAllocationRequests(commonPool shared.Resources) (map[shared.ClientID]shared.Resources, bool) {
-	e.budget -= serviceCharge
+func (e *executive) getAllocationRequests(commonPool shared.Resources) shared.PresidentReturnContent {
 	return e.clientPresident.EvaluateAllocationRequests(e.ResourceRequests, commonPool)
 }
 
-func (e *executive) requestAllocationRequest() {
+//requestRuleProposal asks each island alive for its rule proposal.
```
NGovani: potentially you could as the president which actions it wants to take at the beginning of iigo

internal/server/iigointernal/executive.go
```diff
@@ -48,73 +48,118 @@ func (e *executive) setAllocationRequest(resourceRequests map[shared.ClientID]sh
...
+	if !CheckEnoughInCommonPool(actionCost.ReplyAllocationRequestsActionCost, e.gameState) {
+		return errors.Errorf("Insufficient Budget in common Pool: replyAllocationRequest")
+	}
+	allocationMapReturn := e.getAllocationRequests(commonPool)
+	if allocationMapReturn.ActionTaken && allocationMapReturn.ContentType == shared.PresidentAllocation {
+		if !e.incurServiceCharge(actionCost.ReplyAllocationRequestsActionCost) {
+			return errors.Errorf("Insufficient Budget in common Pool: replyAllocationRequest")
+		}
+		for _, islandID := range aliveIslands {
+			d := shared.CommunicationContent{T: shared.CommunicationInt, IntegerData: int(allocationMapReturn.ResourceMap[islandID])}
```
NGovani: A president might not add a reply to a certain island so its best to make sure there is a value for the corresponding island id in the resourceMap.

This only applies if agents overload the current implementation for EvaluateAllocationRequests

internal/server/iigointernal/executive.go
```diff
@@ -48,73 +48,118 @@ func (e *executive) setAllocationRequest(resourceRequests map[shared.ClientID]sh
...
+	if !CheckEnoughInCommonPool(actionCost.BroadcastTaxationActionCost, e.gameState) {
+		return errors.Errorf("Insufficient Budget in common Pool: broadcastTaxation")
+	}
+	taxMapReturn := e.getTaxMap(islandsResources)
+	if taxMapReturn.ActionTaken && taxMapReturn.ContentType == shared.PresidentTaxation {
+		if !e.incurServiceCharge(actionCost.BroadcastTaxationActionCost) {
+			return errors.Errorf("Insufficient Budget in common Pool: broadcastTaxation")
+		}
+		for _, islandID := range aliveIslands {
+			d := shared.CommunicationContent{T: shared.CommunicationInt, IntegerData: int(taxMapReturn.ResourceMap[islandID])}
```
NGovani: A president might not add a reply to a certain island so its best to make sure there is a value for the corresponding island id in the resourceMap. If not we could potentially use a default tax amount.

internal/server/iigointernal/executive.go
```diff
@@ -48,73 +48,118 @@ func (e *executive) setAllocationRequest(resourceRequests map[shared.ClientID]sh
...
 	var nextSpeaker shared.ClientID
 	electionsettings := e.clientPresident.CallSpeakerElection(e.speakerTurnsInPower, allIslands)
 	if electionsettings.HoldElection {
-		// TODO: deduct the cost of holding an election
+		if !e.incurServiceCharge(actionCost.AppointNextSpeakerActionCost) {
+			return e.PresidentID, errors.Errorf("Insufficient Budget in common Pool: appointNextSpeaker")
+		}
 		election.ProposeElection(baseclient.President, electionsettings.VotingMethod)
 		election.OpenBallot(electionsettings.IslandsToVote)
 		election.Vote(iigoClients)
```
NGovani, line 159: if the next speaker is the same as the last even if they re-win the election should we not increment the speakerTurnsInPower count rather than set it to 0

internal/server/iigointernal/executive.go
```diff
@@ -125,49 +170,47 @@ func (e *executive) appointNextSpeaker(currentSpeaker shared.ClientID, allIsland
...
 
-func getIslandAlive() []float64 {
-	return rules.VariableMap[rules.IslandsAlive].Values
+// incur charges in both budget and commonpool for performing an actions
+// actionID is typically the function name of the action
+// only return error if we can't withdraw from common pool
+func (e *executive) incurServiceCharge(cost shared.Resources) bool {
+	_, ok := WithdrawFromCommonPool(cost, e.gameState)
+	if ok {
+		e.gameState.IIGORolesBudget["president"] -= cost
```
NGovani: is there any check to see if the budget is equal to zero? should a role be able to perform an action if their budget is zero but the common pool still has resources in it?

internal/server/iigointernal/judiciary.go
```diff
@@ -106,7 +98,15 @@ func (j *judiciary) appointNextPresident(currentPresident shared.ClientID, allIs
...
 		nextPresident = currentPresident
 	}
-	return nextPresident
+	return nextPresident, nil
+}
+
+func (j *judiciary) incurServiceCharge(cost shared.Resources) bool {
+	_, ok := WithdrawFromCommonPool(cost, j.gameState)
+	if ok {
+		j.gameState.IIGORolesBudget["judge"] -= cost
```
NGovani: same issue as in the incurServiceCharge for executive branch. Should this not also be in utilities if all 3 roles use similar functions

internal/server/iigointernal/orchestration.go
```diff
@@ -76,64 +91,82 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
...
-	}
-	if !speakerWithdrawSuccess {
-		returnWithdrawnSalariesToCommonPool(g, &executiveBranch, &legislativeBranch, &judicialBranch)
-		return false, "Could not run IIGO since Speaker has no resources to spend"
+	errorJudicial := judicialBranch.sendPresidentSalary()
+	errorLegislative := legislativeBranch.sendJudgeSalary()
+	errorExecutive := executiveBranch.sendSpeakerSalary()
+	// Throw error
+	if errorJudicial != nil || errorLegislative != nil || errorExecutive != nil {
+		return false, "Cannot pay IIGO salary"
```
NGovani: This makes it possible for a president and judge to get their salary even if IIGO doesn't run (because the sendSpeakerSalary failed), since the client states are updated during the sendSalary function calls. Is this the intended behaviour

internal/server/iigointernal/executive.go
```diff
@@ -24,7 +24,7 @@ type executive struct {
 // loadClientPresident checks client pointer is good and if not panics
 func (e *executive) loadClientPresident(clientPresidentPointer roles.President) {
 	if clientPresidentPointer == nil {
-		panic(fmt.Sprintf("Client '%v' has loaded a nil president pointer", e.ID))
+		panic(fmt.Sprintf("Client '%v' has loaded a nil president pointer", e.PresidentID))
```
NGovani, line 29: I know we have an error system for a reason, I also with the current way iigo is setup it would be difficult to return an error. @lhl2617 how bad would this actually be to keep in code?

internal/server/iigointernal/orchestration.go
```diff
@@ -76,64 +91,82 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
...
-	executiveBranch.requestRuleProposal()
-	ruleToVote, ruleSelected := executiveBranch.getRuleForSpeaker()
 
 	// 3 Speaker actions
-	if ruleSelected {
-		legislativeBranch.setRuleToVote(ruleToVote)
-		legislativeBranch.setVotingResult(aliveClientIds)
-		legislativeBranch.announceVotingResult()
+
+	//TODO:- shouldn't updateRules be called here?
```
NGovani, line 162: I agree has this TODO been handled

internal/server/iigointernal/orchestration.go
```diff
@@ -76,64 +91,82 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
...
+		if insufficientBudget != nil {
+			return false, "Common pool resources insufficient for legislativeBranch actions"
+		}
 	}
 
+	// TODO:- at the moment, these are action (and cost resources) but should they?
+	var appointJudgeError, appointSpeakerError, appointPresidentError error
 	// Get new Judge ID
-	g.JudgeID = legislativeBranch.appointNextJudge(g.JudgeID, aliveClientIds)
+	g.JudgeID, appointJudgeError = legislativeBranch.appointNextJudge(g.JudgeID, aliveClientIds)
```
NGovani: If the appointNextJudge function fails the current speaker becomes the new judge, is that intended?

internal/server/iigointernal/orchestration.go
```diff
@@ -76,64 +91,82 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
...
+	if appointJudgeError != nil {
+		return false, "Judge was not apointed by the Speaker. Insufficient budget"
+	}
 	// Get new Speaker ID
-	g.SpeakerID = executiveBranch.appointNextSpeaker(g.SpeakerID, aliveClientIds)
+	g.SpeakerID, appointSpeakerError = executiveBranch.appointNextSpeaker(g.SpeakerID, aliveClientIds)
```
NGovani: Here the president becomes the next speaker if this fails

internal/server/iigointernal/orchestration.go
```diff
@@ -76,64 +91,82 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
...
+	if appointSpeakerError != nil {
+		return false, "Speaker was not apointed by the President. Insufficient budget"
+	}
 	// Get new President ID
-	g.PresidentID = judicialBranch.appointNextPresident(g.PresidentID, aliveClientIds)
+	g.PresidentID, appointPresidentError = judicialBranch.appointNextPresident(g.PresidentID, aliveClientIds)
```
NGovani: Here the judge becomes the next president if this fails

internal/server/iigointernal/utilities.go
```diff
@@ -41,6 +41,12 @@ func WithdrawFromCommonPool(value shared.Resources, gameState *gamestate.GameSta
 	}
 }
 
+func depositIntoClientPrivatePool(value shared.Resources, id shared.ClientID, state *gamestate.GameState) {
```
NGovani, line 44: It would be great if you had logs here. However considering that you'd need to access the server object to log correctly id say its a low priority fix.

NGovani (dismissed):
Overall looks like a good set of additions. Just a few issues need fixing

internal/server/iigointernal/executive.go
Spruudis: the client chooses sequentially which actions it wants to take. So technically the client could decide how to act on the rest of their actions when the first decision is called

internal/server/iigointernal/executive.go
tobhil98: Only change we made was rename ID to PresidentID. We did not implement the panic

internal/server/iigointernal/executive.go
tobhil98: This is an agent implementation issue. If the agent chooses to use more than the allowed budget they are allowed to do so, but they might be punished by the other agents. The function will only return false if you cannot subtract the cost from the common pool.

internal/server/iigointernal/orchestration.go
tobhil98: For the time being this is the intended behavior, yes

internal/server/iigointernal/orchestration.go
tobhil98: Yeah.. Originally this wasn't a problem as returning the error meant that non of the roles would change. This has apparently got lost in one of the merges but should be fixed again now.

internal/server/iigointernal/executive.go
Spruudis: No, we have no design saying a role has a maximum amount of terms. Hence, this should be reset

internal/server/iigointernal/executive.go
Spruudis: Done!

internal/server/iigointernal/executive.go
Spruudis: Good spot, thanks!

Spruudis: Also done!

internal/server/iigointernal/executive.go
```diff
@@ -83,11 +83,13 @@ func (e *executive) broadcastTaxation(islandsResources map[shared.ClientID]share
...
-			d := shared.CommunicationContent{T: shared.CommunicationInt, IntegerData: int(taxMapReturn.ResourceMap[islandID])}
-			data := make(map[shared.CommunicationFieldName]shared.CommunicationContent)
-			data[shared.TaxAmount] = d
-			communicateWithIslands(islandID, shared.TeamIDs[e.PresidentID], data)
+		for islandID, resourceAmount := range taxMapReturn.ResourceMap {
+			if Contains(aliveIslands, islandID) {
+				d := shared.CommunicationContent{T: shared.CommunicationInt, IntegerData: int(resourceAmount)}
+				data := make(map[shared.CommunicationFieldName]shared.CommunicationContent)
+				data[shared.TaxAmount] = d
+				communicateWithIslands(islandID, shared.TeamIDs[e.PresidentID], data)
```
NGovani, line 98: This means that any island not in the resource map is not told of how much tax they need to pay. This is more of a design issue so if that's the intended behaviour then it's fine.

internal/server/iigointernal/executive.go
```diff
@@ -141,11 +143,13 @@ func (e *executive) replyAllocationRequest(commonPool shared.Resources, aliveIsl
...
 			return errors.Errorf("Insufficient Budget in common Pool: replyAllocationRequest")
 		}
-		for _, islandID := range aliveIslands {
-			d := shared.CommunicationContent{T: shared.CommunicationInt, IntegerData: int(allocationMapReturn.ResourceMap[islandID])}
-			data := make(map[shared.CommunicationFieldName]shared.CommunicationContent)
-			data[shared.AllocationAmount] = d
-			communicateWithIslands(islandID, shared.TeamIDs[e.PresidentID], data)
+		for islandID, resourceAmount := range allocationMapReturn.ResourceMap {
+			if Contains(aliveIslands, islandID) {
+				d := shared.CommunicationContent{T: shared.CommunicationInt, IntegerData: int(resourceAmount)}
```
NGovani: If someone ask for an allocation and the president does not respond they are given no update of that. This is fine, it should just be added as a comment in the CommonPoolRequest function in the baseclient

NGovani (commented):
Due to the size of the PR i'd prefer someone else also review this. I'm not gonna approve it but if another reviewer is happy with the changes then I'm also happy for this to be merged in

internal/common/config/config.go
```diff
@@ -73,6 +76,27 @@ type DisasterConfig struct {
 	MagnitudeLambda        float64               // Exponential rate param for disaster magnitude
 }
 
+type IIGOConfig struct {
+	// Executive branch
+	GetRuleForSpeakerActionCost        shared.Resources
```
lhl2617, line 87: these are all variables with verb names that remind me of function names :(

internal/common/shared/iigoreturnstructs.go
```diff
@@ -0,0 +1,50 @@
+package shared
+
+// PresidentReturnContentType is the index of the return content that should be retrieved if the action was taken.
+type PresidentReturnContentType = int
+
+// Enum to specify all the types of possible return values coming from president
+const (
```
lhl2617, line 7: see #150

internal/server/iigointernal/executive.go
lhl2617: #168 author @Neelesh99 approver @mpardalos

params.go
```diff
@@ -153,6 +153,94 @@ var (
 		1,
 		"Exponential rate param for disaster magnitude",
 	)
+
+	// config.IIGOConfig - Executive branch
+	getRuleForSpeakerActionCost = flag.Float64(
```
lhl2617: can we prepend each of these flags with `iigo` so we know these are iigo stuff?

internal/server/server.go
```diff
@@ -69,10 +69,15 @@ func createSOMASServer(
 			DeerPopulation:  foraging.CreateDeerPopulationModel(gameConfig.ForagingConfig.DeerHuntConfig),
 			ForagingHistory: forageHistory,
 			IIGOHistory:     []shared.Accountability{},
-			SpeakerID:       shared.Team1,
-			JudgeID:         shared.Team2,
-			PresidentID:     shared.Team3,
-			CommonPool:      gameConfig.InitialCommonPool,
+			IIGORolesBudget: map[string]shared.Resources{
+				"president": 0,
```
lhl2617: what happened to the iotas?

internal/server/iigointernal/utilities.go
lhl2617: the logger should be passed in. @ me if you need help

internal/server/iigointernal/orchestration.go
```diff
@@ -76,64 +91,86 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
...
+		insufficientBudget = executiveBranch.replyAllocationRequest(g.CommonPool, aliveClientIds)
+	}
+	if insufficientBudget == nil {
+		insufficientBudget = executiveBranch.requestRuleProposal(aliveClientIds)
+	}
+	if insufficientBudget == nil {
+		ruleToVoteReturn, insufficientBudget = executiveBranch.getRuleForSpeaker()
+	}
+	if insufficientBudget != nil {
+		return false, "Common pool resources insufficient for executiveBranch actions"
```
lhl2617: seems like a `try...catch` approach--don't: it's super hard to debug: do you know which function has errored if this returns an error?

internal/server/iigointernal/orchestration.go
```diff
@@ -76,64 +91,86 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
...
-	if !judgeWithdrawSuccess {
-		returnWithdrawnSalariesToCommonPool(g, &executiveBranch, &legislativeBranch, &judicialBranch)
-		return false, "Could not run IIGO since Judge has no resources to spend"
-	}
-	if !speakerWithdrawSuccess {
-		returnWithdrawnSalariesToCommonPool(g, &executiveBranch, &legislativeBranch, &judicialBranch)
-		return false, "Could not run IIGO since Speaker has no resources to spend"
+	errorJudicial := judicialBranch.sendPresidentSalary()
+	errorLegislative := legislativeBranch.sendJudgeSalary()
+	errorExecutive := executiveBranch.sendSpeakerSalary()
```
lhl2617: can you use the `err != nil` pattern?

internal/server/iigointernal/orchestration.go
```diff
@@ -76,64 +91,86 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
...
+	if ruleToVoteReturn.ActionTaken && ruleToVoteReturn.ContentType == shared.PresidentRuleProposal {
+		insufficientBudget := legislativeBranch.setRuleToVote(ruleToVoteReturn.ProposedRule)
+		if insufficientBudget == nil {
+			insufficientBudget = legislativeBranch.setVotingResult(aliveClientIds)
+		}
+		if insufficientBudget == nil {
+			insufficientBudget = legislativeBranch.announceVotingResult()
+		}
+		if insufficientBudget != nil {
+			return false, "Common pool resources insufficient for legislativeBranch actions"
```
lhl2617: see `try...catch` comment.

internal/server/iigointernal/orchestration.go
```diff
@@ -55,14 +56,28 @@ var presidentPointer roles.President = nil
...
+
+	actionCost = gameConf.IIGOConfig
 
 	iigoClients = *clientMap
 
+	// Increments the budget by a constant 100
+	// TODO:- the constant should be retrieved from the rules
+	g.IIGORolesBudget["president"] += 100
+	g.IIGORolesBudget["judge"] += 100
+	g.IIGORolesBudget["speaker"] += 100
```
lhl2617: should be enums

internal/server/iigointernal/orchestration.go
```diff
@@ -2,29 +2,30 @@ package iigointernal
...
 	"github.com/SOMAS2020/SOMAS2020/internal/common/gamestate"
 	"github.com/SOMAS2020/SOMAS2020/internal/common/roles"
-	"github.com/SOMAS2020/SOMAS2020/internal/common/rules"
 	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
 	"github.com/SOMAS2020/SOMAS2020/internal/common/voting"
 )
 
-const serviceCharge = shared.Resources(10)
+// Single source of truth for all action
+var actionCost config.IIGOConfig
```
lhl2617: let's get rid of this global?

internal/server/iigointernal/legislature_test.go
```diff
@@ -124,3 +146,84 @@ func generateRulesTestStores() (map[string]rules.RuleMatrix, map[string]rules.Ru
...
+			returned := tc.bSpeaker.incurServiceCharge(tc.input)
+			commonPool := tc.bSpeaker.gameState.CommonPool
+			presidentBudget := tc.bSpeaker.gameState.IIGORolesBudget["speaker"]
+			if returned != tc.expectedReturn ||
+				commonPool != tc.expectedCommonPool ||
+				presidentBudget != tc.expectedSpeakerBudget {
+				t.Errorf("%v - Failed. Got '%v, %v, %v', but expected '%v, %v, %v'",
+					tc.name, returned, commonPool, presidentBudget,
+					tc.expectedReturn, tc.expectedCommonPool, tc.expectedSpeakerBudget)
+			}
```
lhl2617, line 229: I usually check these one by one, but this is fine, thanks for the tests!!

internal/server/iigointernal/judiciary.go
lhl2617: make an enum instead of using a magic string

internal/server/iigointernal/executive.go
```diff
@@ -125,49 +174,47 @@ func (e *executive) appointNextSpeaker(currentSpeaker shared.ClientID, allIsland
...
 
-func getIslandAlive() []float64 {
-	return rules.VariableMap[rules.IslandsAlive].Values
+// incur charges in both budget and commonpool for performing an actions
+// actionID is typically the function name of the action
+// only return error if we can't withdraw from common pool
+func (e *executive) incurServiceCharge(cost shared.Resources) bool {
+	_, ok := WithdrawFromCommonPool(cost, e.gameState)
+	if ok {
+		e.gameState.IIGORolesBudget["president"] -= cost
```
lhl2617: make enum

lhl2617 (dismissed):
see comments.

params.go
tobhil98: Done

internal/common/config/config.go
tobhil98: Yeah. The verb is the name of the function that uses the config constant. The verb is the action and "actionCost" was to indicate it was the constant.  

If it's not a high priority, we could look at doing the renaming at a later stage if we have time?

internal/server/iigointernal/orchestration.go
tobhil98: If I do `return false, insufficientBudget`. Then the error will indicate which function returned the error. 
I'm not sure what other approaches to use.  We only want to run the next line if the one before was successful. Would you have a return statement after every call?

internal/server/iigointernal/orchestration.go
tobhil98: In the current implementation, we want to pay as many roles as possible. We therefore try paying all three roles and then return an error if that did not succeed.  Are you proposing that we check if a call returns nil before running the next function call?

internal/common/config/config.go
lhl2617: sure

internal/server/iigointernal/orchestration.go
lhl2617: > Are you proposing that we check if a call returns nil before running the next function call?
Yeah

Ok then. can you add this as a comment because this is not the usual error handling behaviour?

internal/server/iigointernal/orchestration.go
lhl2617: > Would you have a return statement after every call? 

you should, if the error is not nil

internal/server/iigointernal/executive.go
Neelesh99: Hey soz just reading this now, the premise of this panic is that if any client for some reason is passing in a nil pointer for this, it is a fundamental problem maybe in the client's implementation. Since we have baseRoles there should be no scenario which has a nil pointer being passed from the client side.

internal/common/baseclient/baseSpeaker.go
```diff
@@ -7,29 +7,48 @@ import (
 type BaseSpeaker struct {
 }
 
-func (s *BaseSpeaker) PayJudge(salary shared.Resources) (shared.Resources, bool) {
-	return salary, true
+// PayJudge is used for paying judge for his service
+func (s *BaseSpeaker) PayJudge(salary shared.Resources) shared.SpeakerReturnContent {
+	return shared.SpeakerReturnContent{
```
Neelesh99, line 12: With these SpeakerReturnContent structs, I'm a little confused, could we achieve the same with more transparent return types. I'm wondering if we might mishandle a function return since they're all of the same type?

internal/common/shared/iigoreturnstructs.go
```diff
@@ -0,0 +1,50 @@
...
+	PresidentAllocation PresidentReturnContentType = iota
+	PresidentTaxation
+	PresidentRuleProposal
+	PresidentSpeakerSalary
+)
+
+// PresidentReturnContent is a general datastructure used for president return type
+type PresidentReturnContent struct {
```
Neelesh99, line 15: As with my previous comment, is this struct necessary. I understand the choice/obligation decision you are trying to emulate but I feel like transparent return types will really help client implementations

internal/common/shared/shared.go
```diff
@@ -26,6 +26,15 @@ const (
 	Team6
 )
 
+const (
```
Neelesh99: I'd thought we already had this in iigoroles.go ?

internal/server/iigointernal/executive.go
```diff
@@ -48,73 +48,122 @@ func (e *executive) setAllocationRequest(resourceRequests map[shared.ClientID]sh
...
-	return e.clientPresident.SetTaxationAmount(islandsResources)
+func (e *executive) getRuleForSpeaker() (shared.PresidentReturnContent, error) {
+	if !CheckEnoughInCommonPool(actionCost.GetRuleForSpeakerActionCost, e.gameState) {
+		return shared.PresidentReturnContent{ContentType: shared.PresidentRuleProposal, ProposedRule: "", ActionTaken: false},
+			errors.Errorf("Insufficient Budget in common Pool: broadcastTaxation")
+	}
+
+	returnRule := e.clientPresident.PickRuleToVote(e.RulesProposals)
+
+	if returnRule.ActionTaken && (returnRule.ContentType == shared.PresidentRuleProposal) {
```
Neelesh99, line 74: Yeah this line highlights the return type shortcomings for me, we shouldn't need to internally be checking things like this.

internal/server/iigointernal/utilities.go
tobhil98: #229 - To be solved here

internal/server/iigointernal/orchestration.go
tobhil98: Thanks, Updated

internal/server/iigointernal/orchestration.go
tobhil98: Done.
Same for the rest of the enum comments

internal/server/iigointernal/orchestration.go
tobhil98: Updated

internal/server/iigointernal/orchestration.go
tobhil98: Done

internal/server/iigointernal/orchestration.go
tobhil98: Done

tobhil98:
@lhl2617 Could you please have another read through when you have time? Still having to fix the tests after updating from main, but most of the changes you requested should now be in place.

internal/server/iigointernal/executive.go
Neelesh99: Could we possibly give these agents like a message with 0 allocations, I feel like it might mess up teams agents if they get nothing at all.

internal/server/iigointernal/executive.go
Neelesh99: I'm unsure of how exactly design intended this, but I can imagine a lot of agent implementations may rely on some form of tax communication even if it is 0.

internal/server/iigointernal/executive.go
```diff
@@ -47,75 +50,105 @@ func (e *executive) setAllocationRequest(resourceRequests map[shared.ClientID]sh
...
+			return false, errors.Errorf("Insufficient Budget in common Pool: replyAllocationRequest")
+		}
 		allocationsMade = true
 		for _, island := range getIslandAlive() {
-			d := shared.CommunicationContent{T: shared.CommunicationInt, IntegerData: int(allocationMap[shared.ClientID(int(island))])}
+			d := shared.CommunicationContent{T: shared.CommunicationInt, IntegerData: int(returnContent.ResourceMap[shared.ClientID(int(island))])}
 			data := make(map[shared.CommunicationFieldName]shared.CommunicationContent)
 			data[shared.AllocationAmount] = d
-			communicateWithIslands(shared.TeamIDs[int(island)], shared.TeamIDs[e.ID], data)
+			communicateWithIslands(shared.TeamIDs[int(island)], shared.TeamIDs[int(e.PresidentID)], data)
```
Neelesh99, line 142: Could we use shared.ClientID(island) instead of shared.TeamIDs[int(island)], the latter could cause an index error and panic rather than be caught by the check in communicateWithIslands

internal/server/iigointernal/executive.go
```diff
@@ -180,3 +219,14 @@ func checkRuleIsValid(ruleName string, rulesCache map[string]rules.RuleMatrix) b
 func getIslandAlive() []float64 {
 	return rules.VariableMap[rules.IslandsAlive].Values
 }
+
+// incur charges in both budget and commonpool for performing an actions
+// actionID is typically the function name of the action
+// only return error if we can't withdraw from common pool
+func (e *executive) incurServiceCharge(cost shared.Resources) bool {
+	_, ok := WithdrawFromCommonPool(cost, e.gameState)
```
Neelesh99, line 232: I know it's a nitpick, but instead of ignoring this value, could we assign and then on line 229 subtract this value. Just to be clear to anyone reading.

internal/server/iigointernal/judiciary.go
```diff
@@ -51,50 +54,38 @@ func (j *judiciary) loadClientJudge(clientJudgePointer roles.Judge) {
...
-	var withdrawAmount, withdrawSuccesful = WithdrawFromCommonPool(presidentSalary, gameState)
-	j.presidentSalary = withdrawAmount
-	return withdrawSuccesful
-}
-
-// sendPresidentSalary sends the president's salary to the president.
-func (j *judiciary) sendPresidentSalary(executiveBranch *executive) {
+// sendPresidentSalary conduct the transaction based on amount from client implementation
+func (j *judiciary) sendPresidentSalary() error {
 	if j.clientJudge != nil {
```
Neelesh99, line 64: If you'd like you can skip this check since we have the nil check in loadClientJudge

internal/server/iigointernal/judiciary.go
```diff
@@ -140,7 +133,15 @@ func (j *judiciary) appointNextPresident(monitoring shared.MonitorResult, curren
 		j.presidentTurnsInPower++
 		nextPresident = currentPresident
 	}
-	return nextPresident
+	return nextPresident, nil
+}
+
+func (j *judiciary) incurServiceCharge(cost shared.Resources) bool {
+	_, ok := WithdrawFromCommonPool(cost, j.gameState)
```
Neelesh99, line 145: Same as with president, could be nice to not throw away this value.

internal/server/iigointernal/legislature.go
```diff
@@ -29,53 +32,58 @@ func (l *legislature) loadClientSpeaker(clientSpeakerPointer roles.Speaker) {
...
-	l.judgeSalary = withdrawAmount
-
-	return withdrawSuccesful
-}
-
-// sendJudgeSalary sets the budget of the Judge.
-func (l *legislature) sendJudgeSalary(judicialBranch *judiciary) {
+// sendJudgeSalary conduct the transaction based on amount from client implementation
+func (l *legislature) sendJudgeSalary() error {
 	if l.clientSpeaker != nil {
```
Neelesh99, line 42: As before feel free to skip this check

internal/server/iigointernal/orchestration.go
```diff
@@ -70,10 +72,26 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
...
 
+	// Increments the budget by a constant 100
+	// TODO:- the constant should be retrieved from the rules
+	g.IIGORolesBudget[shared.President] += 100
+	g.IIGORolesBudget[shared.Judge] += 100
+	g.IIGORolesBudget[shared.Speaker] += 100
+
+	// Pass in gamestate -
+	// So that we don't have to pass gamestate as arguments in every function in roles
+	judicialBranch.gameState = g
```
Neelesh99: Could we maybe have like a judicialBranch.syncWithGame(gamestate, iigoConfig) rather than direct assignment

internal/server/iigointernal/orchestration.go
Neelesh99: Hi, could we just get a confirm on this, I remember that for MVP we specifically put in rollback functions to prevent this behaviour. @Spruudis @ezgimez is this intended?

internal/server/iigointernal/orchestration.go
```diff
@@ -60,7 +62,7 @@ var presidentPointer roles.President = nil
 var iigoClients map[shared.ClientID]baseclient.Client
 
 // RunIIGO runs all iigo function in sequence
-func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.Client) (IIGOSuccessful bool, StatusDescription string) {
+func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.Client, gameConf *config.Config) (IIGOSuccessful bool, StatusDescription string) {
```
Neelesh99, line 65: For a future PR, let's containerise this function, it has become quite a hefty one.

internal/server/iigointernal/orchestration.go
```diff
@@ -143,27 +172,54 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
...
+	}
 
 	variablesToCache = []rules.VariableFieldName{rules.RuleSelected, rules.VoteCalled}
 	valuesToCache = [][]float64{{boolToFloat(ruleSelected)}, {boolToFloat(voteCalled)}}
 	monitoring.addToCache(g.SpeakerID, variablesToCache, valuesToCache)
 
 	speakerMonitored := monitoring.monitorRole(iigoClients[g.JudgeID])
 
+	// TODO:- at the moment, these are action (and cost resources) but should they?
+	var appointJudgeError, appointSpeakerError, appointPresidentError error
```
Neelesh99, line 187: Why not drop these and use the := instantiation?

internal/server/server.go
```diff
@@ -69,10 +69,15 @@ func createSOMASServer(
 			DeerPopulation:  foraging.CreateDeerPopulationModel(gameConfig.ForagingConfig.DeerHuntConfig),
 			ForagingHistory: forageHistory,
 			IIGOHistory:     map[uint][]shared.Accountability{},
-			SpeakerID:       shared.Team1,
-			JudgeID:         shared.Team2,
-			PresidentID:     shared.Team3,
-			CommonPool:      gameConfig.InitialCommonPool,
+			IIGORolesBudget: map[shared.Role]shared.Resources{
```
Neelesh99, line 71: Could we possibly wrap the SpeakerID, JudgeID and PresidentID in the same struct as their budgets, it seems a bit awkward with the budgets grouped together and the ID's separate. Maybe smth like IIGORolesStatus{ID, budget ...}

Neelesh99 (changes requested):
Most of my comments can be safely pushed to another PR. But I'm a little concerned by the RoleReturnContent structs. I think we lose some of the benefits of typing and could lead to dodgy accidental edge cases where the program compiles and is hard to debug. Also, could we get a check on the salary behaviour when we error out on a certain salary? I think we agreed during MVP that if IIGO doesn't run then the salaries should be returned, I don't see why this should change. The PR as a whole is good, I appreciate the amount of testing as well, nice work!

internal/common/shared/shared.go
tobhil98: Yeah.. This was here before the merge from main. It's all using the enum from iigoroles.go now

internal/server/iigointernal/orchestration.go
tobhil98: Because it doesn't let you update the game state in that way. If you add := on appointNextJudge for example you get an error at g.JudgeID

internal/server/iigointernal/orchestration.go
tobhil98: Done

internal/common/baseclient/baseSpeaker.go
tobhil98: Apparently, it was to give the code more structure, as for example before it was returning a boolean with no explanation of what it actually meant if it returned true.

internal/server/iigointernal/orchestration.go
Neelesh99: Oh right I understand, soz didn't notice

internal/server/iigointernal/orchestration.go
Neelesh99: Cool thanks

internal/common/baseclient/baseSpeaker.go
Neelesh99: Could we not have named returns then, I'm not wanting to sound dismissive but I am worried about this disjoint union set up in terms of runtime integrity esp. of functions which clients override, @lhl2617 would you mind weighing in on whether this is ok?

akowal3: if I recall correctly, the previous approach of returning a tuple (value, boolean) was meant to be changed because this return had no meaning. With the return struct, each entry is named and there are checks to check not only the action taken but also that the return was of the expected type (see eg. executive.go, `broadcastTaxation` function)

Neelesh99: That's fair, but looking at these checks (for example line 44 on legislature.go) we simply get a "Cannot perform ...." error which could have been caused by either the type check or !withdrawSuccess. Looking at the return types we have, I'd say we could benefit from named tuple return types and getting the free type checking that the language provides.

Neelesh99: That being said I don't want to block this pr over this, so its fine to leave as is for now.

internal/server/server.go
tobhil98: This doesn't seem to be a very quick fix, as the RoleIDs are used in several places. 
I can try to implement it now if you see it as a high priority, if not I can try fixing it at a later stage in a separate PR?

internal/server/iigointernal/judiciary_test.go
```diff
@@ -263,7 +267,7 @@ func TestInspectHistory(t *testing.T) {
 			}, getBaseEvalResults(shared.TeamIDs)),
 			expectedSuccess: true,
 		},
-		{
+		/*{
```
Neelesh99: Hmmmm sus

tobhil98: Sorry, didn't mean to commit that. It was added alongside the other tests I fixed.

Neelesh99 (approved):
Still not overly happy with the return types, but its good to see the tests passing again. In the interests of time I'm approving for merge, but pls be vigilant of the issues tab if anything comes up.

PR #211, opened 2020-12-31
[INFRA/IIGO] Accountability Cycle Monitoring Result Broadcast to Islands

preetl:
# Summary

Added in a broadcast of the result of monitoring a role in IIGO for islands' opinion formation. Broadcasting the monitoring result is optional and decided by the client.

Part 3 of IIGO  Accountability Cycle.

## Additional Information

This will be extended to influence whether a power transfer is initiated.

## Test Plan

All tests still pass and more will be added once this feature has been fully extended.


QFSW:

internal/server/iigointernal/monitoring.go
```diff
@@ -60,13 +65,28 @@ func (m *monitor) evaluateCache(roleToMonitorID shared.ClientID, ruleStore map[s
...
 	switch roleAccountable.GetID() {
 	case m.speakerID:
-		return m.presidentID, baseclient.President
+		return m.presidentID, shared.President
 	case m.presidentID:
-		return m.judgeID, baseclient.Judge
+		return m.judgeID, shared.Judge
 	default:
-		return m.speakerID, baseclient.Speaker
+		return m.speakerID, shared.Speaker
```
lhl2617: always speaker when the others don't fit?

preetl: Yes, there are only 3 possible cases for calling this func (from President, Speaker or Judge within orchestration.go which is server side) so i didn't know whether it was better to do 3 cases or 2 cases and a default :/

lhl2617: 3 cases + default & error is the best practice :)

preetl: okie dokie, i've changed it!

internal/common/shared/iigocommunications.go
```diff
@@ -49,17 +51,16 @@ type CommunicationContent struct {
 	IntegerData int
 	TextData    string
 	BooleanData bool
+	IIGORole    Role
```
lhl2617, line 54: ahh looks sketch but OK!

lhl2617 (approved):
image

PR #212, opened 2020-12-31
[VIS] Use CSS Animation and app logo for loading

lhl2617:
# Summary

- Previously we used an svg loader which will freeze when the main thread is busy. Use the app logo and css anim to prevent this freezing. See https://www.keycdn.com/blog/animation-performance -- CSS anims that don't repaint/change layout live in the compositor thread

## Test Plan

Freezes no more, but run takes significantly longer. This will be solved by using web workers, that PR awaits

# TODO:- lh: add PR for web worker


QFSW:

Darrekt (approved):
Website go faster

PR #216, opened 2020-12-31
[VIS] Replace slow code-blocks with artifacts with downloads

lhl2617:
# Summary

- Code blocks rendering is slow and makes the site unresponsive
- Change into a table with download links and descriptions
- Add also word-break style to the flags

## Test Plan
- CI Output freezes no more

image

- New run output
image

- Line break

image


QFSW:

QFSW (approved):

PR #221, opened 2020-12-31
[VIS] Auto clear error and display flag type&def

lhl2617:
# Summary

- In a new run, auto clear error upon a successful run
- Show types and defaults for flags

## Test Plan
Errors clear as expected

Types

image


QFSW:

lhl2617:
combine with #223

PR #223, opened 2021-01-01
[VIS] Save output and flags locally

lhl2617:
# Summary

## Storage
- Save flags and output into IndexedDB (localForage) upon a successful run.
- Load these saved items if present on New Run page load
- Clear storage as appropriate during resets.
## New Run
- Clear errors when runs successful
## Flags
- Display type and def properly

## Additional Information

- Fix `<thead><tr>` bug

## Test Plan

- Behaviour described tested successfully.
- Type and def for flag:
  image

### DIFF
--- a/website/src/components/NewRun/utils.ts
+++ b/website/src/components/NewRun/utils.ts
@@ -1,0 +1,71 @@
+import { LOCAL_FLAGS, LOCAL_OUTPUT } from "../../consts/localForage"
+import { Flag, getFlagsFormats, RunGameReturnType, runGame } from "../../wasmAPI"
+import * as localForage from "localforage"
+
+export const loadFlags = async (loadLocal: boolean = true): Promise<Map<string, Flag>> => {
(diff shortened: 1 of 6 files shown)


QFSW:

website/src/components/NewRun/utils.ts
```diff
@@ -0,0 +1,71 @@
+import { LOCAL_FLAGS, LOCAL_OUTPUT } from "../../consts/localForage"
```
olly-larkin: As retrieving from local storage will be useful for individual visualisations as well, would it be better to put this file in a shared utilities folder?

lhl2617: not for now, these utils are reserved for stuff produced by a new run

olly-larkin: Sounds good

olly-larkin (approved):

PR #227, opened 2021-01-02
Disaster mitigation from common pool

JamesTev:
# Summary

See title. 

- [x]  integrate CP contributions/distributions into base client and turn

## Main features
- updated disaster effects to include absolute effects, promotional and CP-mitigated effects
- updated config to include CP-specific disaster params
- Restructured previous work in this branch to use CP in gameState instead of the CP embedded within Env object

## Test Plan

- see TestDisasterEffects. All passing.


QFSW:

internal/server/commonpool.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+}
+
+// islandContribute takes resource donation from individual island into the common-pool
+func (s *SOMASServer) islandContribute(resource shared.Resources, islandID shared.ClientID) {
+	ci := s.gameState.ClientInfos[islandID]
+	ci.Resources = ci.Resources - resource
+	s.gameState.ClientInfos[islandID] = ci
+	s.gameState.CommonPool += shared.Resources(resource)
+}
```
JamesTev: Where would be most logical to include this islandContribute method to request CP contributions from clients?

internal/server/commonpool.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
+)
+
+//islandDistribute distributes resources to island as requested
+func (s *SOMASServer) islandDistribute(resource shared.Resources, islandID shared.ClientID) {
+	ci := s.gameState.ClientInfos[islandID]
+	ci.Resources = ci.Resources + resource
+	s.gameState.ClientInfos[islandID] = ci
+	s.gameState.CommonPool -= resource
+}
```
JamesTev: Where would be most logical to include this method to distribute CP relief to clients?

internal/common/gamestate/gamestate.go
```diff
@@ -43,6 +43,7 @@ type GameState struct {
 // Copy returns a deep copy of the GameState.
 func (g GameState) Copy() GameState {
 	ret := g
+	ret.CommonPool = g.CommonPool
```
JamesTev: Will this work for a float type?

internal/common/disasters/commonpooldefinition.go
```diff
@@ -0,0 +1,49 @@
...
+	tempResource := float64(cpResources)
+	if cpResources >= dConf.CommonpoolThreshold { //exceeds cp threshold
+		newMagnitude = (totalEffect / 2) * dConf.MagnitudeResourceMultiplier // better prep means less effect
+	} else {
+		newMagnitude = totalEffect * dConf.MagnitudeResourceMultiplier // no prep so goodluck
+	}
+
+	//compute remaining effect for each island
+	if newMagnitude <= tempResource { // Case when cp can fully mitigate disaster
+		tempResource = tempResource - newMagnitude
```
JamesTev: will fix all these now

internal/server/commonpool.go
NGovani: The current location is in IIGOTax, while the name suggests its for tax we can repurpose it a make it so there's only a single point in the turn where agents contribute to the CP. Both for tax and for disaster mitigation purposes

internal/server/commonpool.go
NGovani: Currently clients can simply just take from the common pool in IIGOAllocations and if you expect client to just take from the CP in order to get some disaster relief we can add as part of that so that its packaged nicely.

internal/server/commonpool.go
JamesTev: Great, thanks @NGovani! And should we still have two distinct contribution requests within IIGOTax (one for tax, one for CP)?

JamesTev: Since if we do just one request for tax and then allocate (contrib - tax) surplus to CP, how could agents choose to not pay tax but still contribute to CP?

internal/server/commonpool.go
JamesTev: See `applyDisasterEffects()` in `endOfTurn()`. To me it makes most sense to implement disaster effects (including relief based on CP) here since this is a sort of asynchronous event that won't happen on a predictable interval (unlike IIGOAllocations). Thoughts?

internal/server/commonpool.go
NGovani: Hmm thats a good point, I think we can just say that you can by happy accident you pay your tax because you want to contribute to the CP. I doubt an agent which is complicated enough to somehow protest IIGO but still pay to the common pool for the greater good will be created

internal/server/commonpool.go
NGovani: Well from my understanding of the disaster system the common pool is first depleted from in order to mitigate disaster, I guess that acts as a sort of 'relief' and then any agent who wants to take from the common pool for the purposes of surviving a disaster can take it as normal. Potentially we can add a reason as to why the agent is taking from the common pool but that seems low priority

internal/server/commonpool.go
JamesTev: Yeah I agree that it's unlikely, but why would we not just want to have two separate functions to allow for the possibility? It's not like it really complicates things at all? Have already implemented in this PR. Can change back though if there's a strong reason why this wouldn't be a good idea

internal/server/commonpool.go
JamesTev: Hmm I don't think that's how our team implemented it. It's more along the lines of
- get 'raw' effect of storm
- get proportional effect for each island (relative to others)
- reduce effect of each island using CP resources in proportion to effect/damage felt
- deduct islands' resources based on the final CP-mitigated effects (potentially zero deduction if full mitigation)

Would this way of doing it be a problem?

internal/server/commonpool.go
NGovani: Oh if you've implemented it already then that's absolutely fine.  No issues here. The only request I have is could you call the function that prompts islands to contribute to the common pool after IIGOTax but before probeDisaster in runEndOfTurn. That just means that the order of actions regarding the common pool stays the same

JamesTev: Sure thing 👍

internal/server/commonpool.go
NGovani: I think this way of calculating disaster effects is fine. I think we actually agree, I just kinda muddled my words. In terms of directly giving resources to clients from the common pool let's leave it to just happen at IIGOAllocations

internal/common/baseclient/iigo.go
```diff
@@ -53,6 +53,13 @@ func (c *BaseClient) GetTaxContribution() shared.Resources {
 	return 0
 }
 
+// GetCommonPoolContribution requests a client resource contribution towards the common pool.
+// This is used for disaster mitigation in the event of a disaster.
+// OPTIONAL: Use this method to contribue towards CP
+func (c *BaseClient) GetCommonPoolContribution() shared.Resources {
+	return 0
+}
```
preetl: I think that this is unnecessary, there is already a function called GetTaxContribution() which covers contributions to the common pool. As I understand it, that contribution from an island should cover everything (apart from sanction payments).

internal/common/gamestate/gamestate.go
preetl: I think you don't need this line. CommonPool will get copied over as it should in the line ret:=g as it is a variable. This doesn't work for reference types (pointers) which is why ClientInfos etc. has to be copied over in this way.

JamesTev: True, thanks Preet

internal/common/baseclient/iigo.go
JamesTev: Yeah I was aware of this. But thought having a separate function offers more flexibility. Otherwise agents can't not pay tax but still contribute to CP. See my first discussion with Naim above. Open to unifying if there's a strong reason against separate methods but it makes more sense to me tbh - especially since CP contributions are conceptually different to paying tax.

preetl: Maybe we should check with design on this as I think they conceptually are the same (or similar). 

The point of the tax is to make sure that there is enough in the common pool to pay for IIGO on the next turn and to mitigate a disaster. The President sets a minimum contribution but islands are free to pay more. If there are 2 separate functions, this implies that an island can contribute to the common pool without paying tax but tax is defined as a minimum contribution so they have in fact paid the tax. Hope that makes sense!

JamesTev: What if they want to contribute an amount to CP that is below the min tax amount set by President?

Spruudis: To me they don't seem to be conceptually different, a tax is just a rule about the minimum contribution an agent should make. It is always fine to contribute more. With this in mind your argument about agents that can't not pay tax but still contribute to the CP does not make sense - if they contribute enough they have paid the tax.

preetl: > What if they want to contribute an amount to CP that is below the min tax amount set by President?

That's fine, it's entirely the agent's choice. They just might get caught and sanctioned by the Judge at the next IIGO session.

JamesTev: @Spruudis not sure I agree. When would you ever pay more tax than required in a practical case? Also, I was referring specifically to the case where they _don't_ contribute enough to cover base tax amount. Anyway, this is just going to take too long to debate or change so I'm just going to merge the CP contributions into tax as you've suggested.

internal/server/iigo.go
```diff
@@ -30,8 +30,8 @@ func (s *SOMASServer) updateIIGOTurnHistory(clientID shared.ClientID, pairs []ru
 }
 
 func (s *SOMASServer) runIIGOTax() error {
-	s.logf("start runIIGOTax")
```
JamesTev, line 33: @preetl Please just verify that this is what you had in mind i.t.o. the CP contributions integrated into tax requests. Is there anything else that should be added for CP contributions?

preetl: Looks great! Thanks for making the changes :)

preetl (approved):
I'll make an issue re the naming of tax/contirbution. Thanks James!

params.go
```diff
@@ -12,12 +12,12 @@ var (
...
 	maxSeasons = flag.Uint(
 		"maxSeasons",
-		100,
+		10,
 		"The maximum number of 1-indexed seasons to run the game.",
 	)
 	maxTurns = flag.Uint(
 		"maxTurns",
-		100,
+		10,
```
lhl2617, line 20: Why was this reduced?

PR #228, opened 2021-01-02
[VIS] Role visualisation

olly-larkin:
# Summary

Added role visualisation. Updates for new runs (currently no change seen as voting does not occur).

image

## Additional Information

Currently using old version of recharts as newest has bug which causes compilation errors.

## Test Plan

If disabling using new runs and modify initial `output.json`, changes can be seen after refreshing the page.

image

### DIFF
--- a/website/src/components/Visualisations/Roles/Roles.module.css
+++ b/website/src/components/Visualisations/Roles/Roles.module.css
@@ -1,3 +1,18 @@
+.text {
+    font-size: 2em;
+    font-weight: 600;
+}
+
+.root {
+    margin: 0;
+    width: 50%;
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    -ms-transform: translate(-50%, -50%);
+    transform: translate(-50%, -50%);
+}
+
 .customTooltip {
     background-color: rgba(255, 255, 255, 0.95);
     padding: 1rem;
--- a/website/src/components/Visualisations/Roles/Roles.module.css
+++ b/website/src/components/Visualisations/Roles/Roles.module.css
@@ -1,33 +1,14 @@
-.text {
-    font-size: 2em;
-    font-weight: 600;
+.customTooltip {
+    background-color: rgba(255, 255, 255, 0.95);
+    padding: 1rem;
+    border-radius: 0.2rem;
 }
 
-.root {
-    margin: 0;
-    position: absolute;
-    top: 50%;
-    left: 50%;
-    -ms-transform: translate(-50%, -50%);
-    transform: translate(-50%, -50%);
+.customTooltip p.label {
+    text-align: left;
+    font-weight: bold;
 }
 
-.appLogo {
-    height: 40vmin;
-    pointer-events: none;
-}
-
-@media (prefers-reduced-motion: no-preference) {
-    .appLogo {
-        animation: App-logo-spin infinite 20s linear;
-    }
-}
-
-@keyframes App-logo-spin {
-    from {
-        transform: rotate(0deg);
-    }
-    to {
-        transform: rotate(360deg);
-    }
+.customTooltip p.content {
+    text-align: left;
 }
--- a/website/src/components/Visualisations/Roles/Roles.tsx
+++ b/website/src/components/Visualisations/Roles/Roles.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import styles from "./Roles.module.css";
 import {
     BarChart,
@@ -10,12 +10,8 @@ import {
     ResponsiveContainer,
 } from "recharts";
 import { TurnsInRoles, ProcessedRoleData } from "./Util/RoleTypes";
-import { getProcessedRoleData } from "./Util/ProcessedRoleData";
-
-const presidentColor = "#00bbf9";
-const judgeColor = "#fee440";
-const speakerColor = "#f15bb5";
-const noneColor = "#b2bec3";
+import { getProcessedCIRoleData, processRoleData } from "./Util/ProcessedRoleData";
+import { loadLocalOutput } from "../../NewRun/utils";
 
 type CustomTooltipProps = {
     active: boolean;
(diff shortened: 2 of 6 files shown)


QFSW:

lhl2617:
image
nice

website/src/components/Visualisations/Roles/Roles.module.css
```diff
@@ -5,29 +5,25 @@
 
 .root {
     margin: 0;
+    width: 50%;
```
lhl2617, line 8: this might not look good on mobile

lhl2617: but that's really a stretch goal, so nvm. you can however try `min-width: 50%` and see how it looks like

website/src/components/Visualisations/Roles/Roles.tsx
```diff
@@ -1,14 +1,174 @@
...
+    Bar,
+    XAxis,
+    YAxis,
+    Tooltip,
+    Legend,
+    ResponsiveContainer,
+} from "recharts";
+import { TurnsInRoles, ProcessedRoleData } from "./Util/RoleTypes";
+import { getProcessedCIRoleData, processRoleData } from "./Util/ProcessedRoleData";
+import { loadLocalOutput } from "../../NewRun/utils";
```
lhl2617: would need to refactor after #230 goes in

website/src/components/Visualisations/Roles/Roles.tsx
```diff
@@ -1,14 +1,174 @@
...
+            turnsInRoles.judge +
+            turnsInRoles.speaker +
+            turnsInRoles.none;
+
+        return (
+            <div className={styles.customTooltip}>
+                <p className={styles.label}>{label}</p>
+                <p className={styles.content}>
+                    Turns as President: {turnsInRoles.president} (
+                    {((turnsInRoles.president * 100) / turns).toFixed(1)}%)
```
lhl2617: what is this 100 magik numba?

website/src/components/Visualisations/Roles/Roles.tsx
```diff
@@ -1,14 +1,174 @@
...
+                </p>
```
lhl2617: fine but you can use some sik tricks to map from `[president, judge, speaker, none].map(....// sik trick here)`

website/src/components/Visualisations/Roles/Roles.tsx
```diff
@@ -1,14 +1,174 @@
...
+};
 
 const Roles = () => {
-  return (
-    <div className={styles.root}>
-      <img src={logo} className={styles.appLogo} alt="logo" />
-      <p className={styles.text}>Role Visualisation</p>
-    </div>
-  );
+    const presidentColor = "#00bbf9";
```
lhl2617: mmm  lots of hardcoded roles so might be cumbersome if we add roles, but we won't--so this is fine.

website/src/components/Visualisations/Roles/Roles.tsx
```diff
@@ -1,14 +1,174 @@
...
+            setData(processRoleData(localOutput.output));
+        }
+    };
+
+    return (
+        <div className={styles.root}>
+            <ResponsiveContainer height={460} width="100%">
+                <BarChart data={data} layout="vertical" margin={{ bottom: 30 }}>
+                    <YAxis type="category" dataKey="name" />
+                    <XAxis type="number" />
```
lhl2617: label turns?

website/src/components/Visualisations/Roles/Util/RoleTypes.ts
```diff
@@ -0,0 +1,101 @@
...
+    constructor(
+        president?: number,
+        judge?: number,
+        speaker?: number,
+        none?: number
+    ) {
+        this.president = president || 0;
+        this.judge = judge || 0;
+        this.speaker = speaker || 0;
+        this.none = none || 0;
```
lhl2617, line 18: is 0 Team1's value?

lhl2617: ah - ok, these are num turns in roles.

website/src/components/Visualisations/Roles/Util/RoleTypes.ts
```diff
@@ -0,0 +1,101 @@
...
+    name: string;
+    roles: TurnsInRoles[];
+
+    constructor(name: string, roles: TurnsInRoles[] = [new TurnsInRoles()]) {
+        this.name = name;
+        this.roles = roles;
+    }
+
+    increment(role: Role) {
+        if (this.roles[this.roles.length - 1].get(role) === 0) {
```
lhl2617: ~vibe~ bounds check

website/src/components/Visualisations/Roles/Util/RoleTypes.ts
```diff
@@ -0,0 +1,101 @@
...
+            const newTurnsInRoles = new TurnsInRoles();
+            newTurnsInRoles.increment(role);
+            this.roles.push(newTurnsInRoles);
+        } else {
+            this.roles[this.roles.length - 1].increment(role);
```
lhl2617: ~vibe~ bounds check

lhl2617 (changes requested):
Looks really good to me, good job. 

Let's merge in #230 and #231 first

website/src/components/Visualisations/Roles/Roles.tsx
QFSW: Looks like decimal to percentage conversion in which case the magic number is fine

website/src/components/Visualisations/Roles/Util/RoleTypes.ts
lhl2617: use `??` btw -- i think 0 || 1 -> 1 (not an issue here but just good pratice)

website/src/components/Visualisations/Roles/Roles.module.css
olly-larkin: Going to leave this for now as I can't see how this graph would work on mobile without rotating (at which point just rotate the device)

lhl2617 (approved):
this boi clean
image

Issue #224, opened 2021-01-01
[IIGO] Pass server pointer as logger to use in `iigointernal`


lhl2617:
https://play.golang.org/p/GV5GQ-HJMeT

lhl2617:
resolved by #229

PR #229, opened 2021-01-02
[INFRA/IIGO] RE: Issue 224 - Logging in IIGO 

preetl:
# Summary

EDIT: Logs added for IIGO. I haven't logged everything as quite a lot of actions happen in IIGO but we can change this if necessary.

Just a test to work out how this would be possible. @lhl2617 would you mind telling me if this fits what you had described in issue #224 ? I'm not sure if it's the right way to do it but it does seem to work. Will add in more helpful logging statements if it's correct.

## Additional Information

N/A

## Test Plan

N/A


QFSW:

internal/server/iigo.go
```diff
@@ -13,7 +13,7 @@ func (s *SOMASServer) runIIGO() error {
 
 	nonDead := getNonDeadClientIDs(s.gameState.ClientInfos)
 	updateAliveIslands(nonDead)
-	iigoSuccessful, iigoStatus := iigointernal.RunIIGO(&s.gameState, &s.clientMap)
+	iigoSuccessful, iigoStatus := iigointernal.RunIIGO(s.iigoServer, &s.gameState, &s.clientMap)
```
lhl2617: you can just pass in `s.Logf` i think

internal/server/iigointernal/orchestration.go
```diff
@@ -9,6 +9,14 @@ import (
 	"github.com/SOMAS2020/SOMAS2020/internal/common/voting"
 )
 
+type IIGOServer struct {
+	IIGOLogging func(format string, a ...interface{})
```
lhl2617: logger is a nicer name :) not need to export as well

lhl2617:
yes, this is correct. you should now add loggers to your `judiciary` etc. struct that uses the server's logger.

even better that you can print in the format
`[SERVER][JUDICIARY]: ...` for example

internal/server/iigo.go
preetl: oh so no need to make the iigoserver struct ? lol

preetl: or do you mean s.iigoServer.logf ?

lhl2617: `s.logf` is a logger function you can pass in

lhl2617:
image

indeed

lhl2617:
the logger type is actually same for all loggers, so perhaps we can move it into shared.

NGovani:
I think he meant that since the three roles perform transactions such as taking from the common pool and giving resources to each other, these transactions should be logged correctly. In order to do that the role objects need access to the logger

lhl2617:
See #255

lhl2617:
is this active? @preetl

preetl:
> is this active? @preetl

I'm just adding in logging statements, then I'll make it ready for review!

internal/server/iigointernal/judiciary.go
```diff
@@ -235,6 +241,7 @@ func (j *judiciary) scoreIslandTransgressions(transgressions map[shared.ClientID
 			} else {
 				totalIslandTurnScore += roles.IIGOSanctionScore(1)
 			}
+			j.Logf("Rule: %v, broken by %v", ruleBroken, islandID)
```
Neelesh99: @ramon15749 : What's this inconsistency with the colons @preetl ?

Neelesh99 (changes requested):
Almost there, ramon had a comment for ya

Neelesh99 (approved):
LGTM

PR #213, opened 2020-12-31
fixed vis nav routes with wrapper for dropddown

yp717:
# Summary

fixed vis nav routes with wrapper for dropddown

## Additional Information

Auxiliary information goes here.

## Test Plan

- Add information on how you tested your changes.

PR #230, opened 2021-01-02
[VIS] Rebuild visualisations interface

lhl2617:
# Summary

- Remove dropdown for vis
- Add vis upload json page
  - image
- Add vis inner navbar
  - image
- Errors in vis when unsupported file:
  - image
  - image
- Warning when git hash mismatch
  - image
- Uploaded JSON stored in `localForage`
  - This will be loaded when the vis page is visited. We can then visualise "New Run" and "CI Output"
- Pass `output` data to each of the Vis props

## Additional Information
- Revert Navbar.Brand Link-wrapping in #213 by @Darrekt -- this breaks link highlighting. It's fine to reload the page.
- Make `OutputJSONType` a constant instead of calling `typeof` everyime needed.

## Test Plan

See screenshots.

### DIFF
--- a/website/src/components/Visualisations/Visualisations.tsx
+++ b/website/src/components/Visualisations/Visualisations.tsx
@@ -1,6 +1,152 @@
-export { default as Game } from "./Game/Game";
-export { default as Resources } from "./Resources/Resources";
-export { default as Roles } from "./Roles/Roles";
-export { default as IIGO } from "./IIGO/IIGO";
-export { default as IITO } from "./IITO/IITO";
-export { default as IIFO } from "./IIFO/IIFO";
+import React, { useEffect, useState } from "react"
+import VisualisationsNavbar from "./VisualisationsNavbar"
+import { Button, Alert } from "react-bootstrap"
+import { useHistory, Route, Switch } from "react-router-dom";
+import { gamevisualisation, visualisations, iifovisualisation, iigovisualisation, iitovisualisation, resourcesvisualisation, rolesvisualisation } from "../../consts/paths";
+import { OutputJSONType } from "../../consts/types";
+import { GitHash } from "../../consts/info";
+import { initialLoadingState, useLoadingState } from "../../contexts/loadingState";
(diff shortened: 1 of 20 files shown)


QFSW:

website/src/components/Visualisations/Visualisations.tsx
```diff
@@ -1,6 +1,154 @@
...
+      }
+    }
+    catch (err) {
+      // if error, just assume not stored at all
+      console.error(err)
+    }
+    setLoading(initialLoadingState)
+  }
+
+  useEffect(() => {
```
olly-larkin, line 45: Do we need 2 separate calls to `useEffect`?

olly-larkin: Or is having multiple calls still normal practice?

website/src/components/Visualisations/Visualisations.tsx
```diff
@@ -1,6 +1,154 @@
...
+      {
+        warning &&
+        <Alert variant="warning" onClose={() => setWarning(undefined)} dismissible className="custom-alert">
+          <Alert.Heading>Rough seas ahead!</Alert.Heading>
+          <p>{warning}</p>
+        </Alert>
+      }
+      {
+        output ?
+          <>
```
olly-larkin: Is a fragment needed if it only has 1 child?

olly-larkin (changes requested):
Only a couple small comments - looks good otherwise

website/src/components/Visualisations/Visualisations.tsx
lhl2617: note the sensitivity list of `useEffect`--the above one is empty and means it runs on component first mount. The second one has `[output]`--it fires everytime output is changed.

website/src/components/Visualisations/Visualisations.tsx
lhl2617: good spot

olly-larkin (approved):
lgtm

PR #232, opened 2021-01-02
[VIS] Fix Navbar.Brand and tr index

lhl2617:
# Summary

- Revert Navbar.Brand Link-wrapping in #213 by @Darrekt -- this breaks NavLink highlighting. It's fine to reload the page.
- Add id to tr in Artifacts, it is a list and we need unique ids for elements in a list.

## Test Plan
- NavLink fixed.
- No more console errors complaining for tr id.

PR #233, opened 2021-01-02
[VIS] Working interactive resource graph using output json

augusta100:
# Summary
Resources graph with selectable teams and turns. Shows end of seasons.

## Additional Information
Need to change data source later after #230 merged

## Test Plan

- idk how to test stuff sorry :/

Screenshot 2021-01-02 at 18 22 54


lhl2617:
we're not doing tests on vis so it's fine, if you're convinced it works well with a wide range of inputs we good.

website/tsconfig.json
```diff
@@ -18,7 +18,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "noImplicitAny": false
```
lhl2617, line 22: hmmm, this isn't good, can you remove and solve the problems where you need implicit any's?

website/src/components/visualisation/Resources/Resources.tsx
```diff
@@ -1,14 +1,14 @@
...
+import LineRechartComponent from './LineGraph';
 
 const Resources = () => {
   return (
     <div className={styles.root}>
-      <img src={logo} className={styles.appLogo} alt="logo" />
-      <p className={styles.text}>
-        Self-Organising Multi-Agent Systems 2020
-        </p>
+      <h1>Resources over Time </h1>
```
lhl2617:
```suggestion
      <h1>Resources over Time</h1>
```

website/src/components/visualisation/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,147 @@
...
+            </ResponsiveContainer >
+        )
+    };
+}
+
+
+export default LineRechartComponent;
+
+
+
```
lhl2617: if you're using vscode `cmd+shift+f` should autoformat and remove these lines

website/tsconfig.json
augusta100: I'll have a look

website/src/components/visualisation/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,147 @@
+import React from "react";
+import { Brush, ResponsiveContainer, LineChart, Line, YAxis, XAxis, CartesianGrid, Tooltip, Legend, Surface, Symbols, ReferenceLine } from '../../../../node_modules/recharts';
```
lhl2617:
```suggestion
import { Brush, ResponsiveContainer, LineChart, Line, YAxis, XAxis, CartesianGrid, Tooltip, Legend, Surface, Symbols, ReferenceLine } from 'recharts';
```
don't need to specify the direct path if it's a node module :)

website/src/components/visualisation/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,147 @@
...
+import outputJSON from '../../../output/output.json';
+import _ from "../../../../node_modules/lodash";
```
lhl2617:
```suggestion
import _ from "lodash";
```

website/src/components/visualisation/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+
+class LineRechartComponent extends React.Component<{}, any> {
```
lhl2617: haven't seen class components in some time... but this actually looks like a neat use case for class, I'm fine with this. We usually use hooks as per other files.

(I started react with class components, and still have a soft spot for them!)

website/src/components/visualisation/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,147 @@
...
+        if(_.includes(this.state.disabled, dataKey)) {
+            this.setState({
+                disabled: this.state.disabled.filter(obj => obj !== dataKey)
+            });
+        }
+        else {
+            this.setState({
+                disabled:this.state.disabled.concat(dataKey)
+            });
+        }
```
lhl2617: Don't think you need lodash here
```
// nice object decomposition to make it cleaner
const { disabled } = this state
if (disabled.includes(dataKey)) {
   this.setState( ... )
}
else {
   this.setState( ... )
}
```

website/src/components/visualisation/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,145 @@
...
+    };
+
+    renderCustomizedLegend = ({ payload }) => {
+        return(
+            <div className="customized-legend">
+                {payload.map(entry => {
+                    const { dataKey, colour } = entry;
+                    const active = _.includes(this.state.disabled, dataKey);
```
lhl2617: similarly, no lodash needed

website/src/components/visualisation/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,145 @@
...
+        }
+        return seasonEnds;
+    };
+
+    render() {
+        return (
+            <ResponsiveContainer height={330} width = "95%">
+                <LineChart data={this.state.chartData.GameStates}
+                    margin={{ top: 5, right: 30, left: 20, bottom: 5 }}>
+                    {_.toPairs(this.state.lineColours)
```
lhl2617: hmm -- does this look indented ok to you? also it looks to have a lot of logic, so we could make this a class fn

lhl2617: @olly-larkin i think your diff can benefit from` _.toPairs`, seems neat

lhl2617:
good job, overall looks ok to me.
will approve later after we merge in #230 - #232.

you also need to update from main

website/src/components/visualisation/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,145 @@
...
+class LineRechartComponent extends React.Component<{}, any> {
+
+    //data = outputJSON;
+
+    constructor(props){
+        super(props);
+        this.state = {
+            disabled: [],
+            lineColours: {
+                team1: "#0095FF",
```
lhl2617: #234 merged, so we don't need to hardcode these now. for colours it's a bit of a pain but just put in some random colours after?

lhl2617: (I'd personally write soemthing that generates colours based on a spectrum, but that's too much work. you can just make it loop around 6 colours as well if it's labelled ok.

website/src/components/visualisation/Resources/LineGraph.tsx
ahm116: Do you mean a class function for LineChart? Or a class function the Line lines in the chart?

ahm116: And the tabbing in those four line 98-101 are right but the tabbing below was wrong. I did the autoformat  thing so should all be tabbed properly now

augusta100: > hmm -- does this look indented ok to you? also it looks to have a lot of logic, so we could make this a class fn

The indenting is right for 98-101 but was wrong below that - I did the autoformat thing so should all be fine now. Do you mean a class fn for LineChart or just the Line lines?

lhl2617: the lines from 101 onwards

website/tsconfig.json
Darrekt: Feel free to ping us on discord if you need advice on how to type stuff, it's a bit tricky at first.

website/src/components/visualisation/Resources/LineGraph.tsx
Darrekt: @ahm116 might be `ctrl (command) + shift + i`, I think lh has a custom keybind.

Additionally, some autoformatters (prettier) think that rogue newlines are user-intended and leave them alone. It's good to check manually for this. In general for typescript, 1 line between every code block. New classes are usually in a separate file, so no need to do the "2 lines between classes, 1 between functions" thing that python does.

website/src/components/Visualisations/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,143 @@
+import React from "react";
+import { Brush, ResponsiveContainer, LineChart, Line, YAxis, XAxis, CartesianGrid, Tooltip, Legend, Surface, Symbols, ReferenceLine } from 'recharts';
```
Darrekt: Large imports like this usually get split into new lines, or get namespaced. Are you using Prettier?

website/src/components/Visualisations/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,143 @@
...
+        }
+    };
+
+    renderCustomizedLegend = ({ payload }) => {
+        return (
+            <div className="customized-legend">
+                {payload.map(entry => {
+                    const { dataKey, colour } = entry;
+                    const { disabled } = this.state;
+                    const active = disabled.includes(dataKey);
```
Darrekt:
```suggestion
                    const active = this.state.disabled.includes(dataKey);
```

Darrekt: Doesn't seem like you use `disabled` anywhere else

website/src/components/Visualisations/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,143 @@
...
+                team5: "ClientInfos.Team5.Resources",
+                team6: "ClientInfos.Team6.Resources",
+            },
+            chartData: outputJSON
+        };
+    }
+
+    handleClick = dataKey => {
+        const { disabled } = this.state;
+        if (disabled.includes(dataKey)) {
```
Darrekt, line 35:
```suggestion
        if (this.state.disabled.includes(dataKey)) {
```

Darrekt: In fact, you still use `this.state.disabled` below.

website/src/components/Visualisations/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    )
+                })}
+            </div>
+        );
+    };
+
+    getSeasonEnds() {
+        var seasonEnds: any[] = [];
+        let i;
+        for (i in this.state.chartData.GameStates) {
```
Darrekt, line 100: We have `forEach` in javascript on arrays, give it a try!

Darrekt (changes requested):
Looks good, I just have a bunch of pedantic remarks. See comments.

You haven't hooked this up to the actual navigator, would love to see that and try running it on my local branch. (we can help you with this, just let us know).

website/src/components/Visualisations/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,143 @@
...
+        return (
+            <div className="customized-legend">
+                {payload.map(entry => {
+                    const { dataKey, colour } = entry;
+                    const { disabled } = this.state;
+                    const active = disabled.includes(dataKey);
+                    const style = {
+                        marginRight: 10,
+                        colour: active ? "#AAA" : "#000"
+                    };
```
Darrekt: Not important, but would like to discuss for fun:

I usually like declaring constants like this at the top, so in this case between l47 & l48. I feel it makes code a bit clearer that the constants aren't part of the imperative looping logic.

website/src/components/Visualisations/Resources/LineGraph.tsx
```diff
@@ -0,0 +1,143 @@
...
+        if (disabled.includes(dataKey)) {
+            this.setState({
+                disabled: this.state.disabled.filter(obj => obj !== dataKey)
+            });
+        }
+        else {
+            this.setState({
+                disabled: this.state.disabled.concat(dataKey)
+            });
+        }
```
Darrekt:
```suggestion
this.setState({
      disabled: this.state.disabled.includes(dataKey)
        ? this.state.disabled.filter((obj) => obj !== dataKey)
        : this.state.disabled.concat(dataKey),
    });
```

My indentation is broken but you get the point

augusta100 (commented):
Added all these changes

Darrekt (approved):
S T A M P

PR #235, opened 2021-01-02
[INFRA/IIGO] Broadcasting sanction thresholds

Neelesh99:
# Summary

Since we are now allowing the Judge to load custom sanction thresholds, it's important that islands know about them.

## Test Plan

Tests added for broadcast generation.

### DIFF
--- a/internal/common/shared/iigocommunications.go
+++ b/internal/common/shared/iigocommunications.go
@@ -91,6 +91,8 @@ func (c CommunicationFieldName) String() string {
 		"SanctionAmount",
 		"RoleMonitored",
 		"MonitoringResult",
+		"IIGOSanctionTier",
+		"IIGOSanctionScore",
 	}
 	if c >= 0 && int(c) < len(strs) {
 		return strs[c]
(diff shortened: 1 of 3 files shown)


QFSW:

preetl:
SO PROUD OF YOU FOR MAKING A LIL PR!!!

internal/common/shared/iigocommunications.go
```diff
@@ -71,6 +71,8 @@ const (
 	SanctionAmount
 	RoleMonitored
 	MonitoringResult
+	IIGOSanctionTier
+	IIGOSanctionScore
```
lhl2617, line 75: did you update the String() function

Neelesh99: Crap, just realised this Enum hasn't been moved over to that model. Let me add that.

lhl2617 (approved):
image

thanks for the smol pr

pre-approving, feel free to merge after fixing String() for enum

internal/common/shared/iigocommunications.go
Neelesh99: Nvm I was confused, added to the .String() function

PR #204, opened 2020-12-30
Change order End of Turn

E-Mac27:
# Summary
Changing order of End of Turn as stated in the design meeting

@yp717 @Spruudis

PR #236, opened 2021-01-02
IITO: Allow clients to control how much to give in GiftTransactions

E-Mac27:
# Summary
I messed up in #204 and this commit was reverted.

Clients previously was forced to give the amount they promised initially. This isn't quite inline with IITO design. Since giftsTransactions execute at the end of turn, clients state have changed.

Now they are able to specify if they want to fulfil the amount they initially promised or change it.

## Test Plan
Tests all pass.


QFSW:

NGovani (approved):

PR #238, opened 2021-01-02
[VIS] Add fake wait in visualise button for stability

lhl2617:
As title. Previously it becomes a bit unstable when the run is a long 'un

## Test plan

Not unstable anymore


QFSW:

NGovani (approved):

PR #243, opened 2021-01-02
[VIS] Fix Navbar.Brand, finally

lhl2617:
# Summary

Right now it goes to `/` which is not the correct URL.

## Test Plan
Goes to home page properly, and unhighlights other links on navbar


QFSW:

Darrekt (approved):

PR #247, opened 2021-01-03
[INFRA/RULES] Layered rules implemented for IIGO conditional rules

Neelesh99:
# Summary

Newly conceptualised rules for IIGO conditionals requires an extension to the rules infrastructure. There should be no regression impacts and this is tested. Linked rules allow for a parent rule to conditionally execute a child rule (kind of like an if statement).

## Test Plan

Added tests


QFSW:

internal/common/rules/rulematrix.go
```diff
@@ -4,10 +4,27 @@ import (
...
+const (
+	ParentFailAutoRulePass LinkTypeOption = iota
+	NoLink
+)
+
+// RuleLink provides a containerised package for all linked rules
+type RuleLink struct {
+	Linked     bool
+	LinkType   LinkTypeOption
+	LinkedRule string
```
Spruudis, line 21: Should this contain a RuleMatrix instead? How would you go about changing the linked rule? Do we still have a rulesInPlay matrix and if so, do both of the linked and the mother rule need to be in play for it to be evaluated properly?

Neelesh99: I think we should keep it as a string reference to the rule name, to ensure minimal regression impacts on voting and evaluation. Yes both the linked and mother rule needs to be in play, this makes sense right, we can vote to modify or remove either independent from each other. We don't need to modify any of our procedures but we still get the intended effect.

internal/common/rules/basicruleevaluator.go
```diff
@@ -167,3 +167,32 @@ func BasicRealValuedRuleEvaluator(ruleName string) (bool, float64, error) {
...
+			if link.LinkType == ParentFailAutoRulePass {
+				childRule := link.LinkedRule
+				parentPass, parentErr := BasicBooleanRuleEvaluator(ruleName)
+				if parentErr != nil {
+					return false, errors.Errorf("Parent Rule errored out wiht : %v", parentErr)
+				}
+				if !parentPass {
+					return true, nil
+				}
+				childPass, childErr := BasicBooleanRuleEvaluator(childRule)
```
Spruudis: Here we evaluate the child rule even if it is not in play, is that right or am I misunderstanding?

Neelesh99: Yes, all the rule evaluators feed off the available rules cache to allow for islands to look at prospective rules. It's up to us to check whether the rules are in play before we call this.

internal/common/rules/basicruleevaluator_test.go
```diff
@@ -43,12 +54,38 @@ func registerNewRealValuedRule(t *testing.T) {
...
+
+	v := []float64{1, 0, 0, 0, -4, 0, -1, -1, 0, 2, 0, 0, 0, 0, 2, 0, 0, 1, 0, -1}
+	CoreMatrix := mat.NewDense(4, 5, v)
+	aux := []float64{1, 1, 3, 0}
+	AuxiliaryVector := mat.NewVecDense(4, aux)
+
+	_, ruleError := RegisterNewRule(name, reqVar, *CoreMatrix, *AuxiliaryVector, false, RuleLink{
+		Linked:     true,
+		LinkType:   ParentFailAutoRulePass,
+		LinkedRule: "Kinda Complicated Rule",
```
NGovani, line 86: Where is this rule defined

internal/common/rules/rulematrix.go
```diff
@@ -4,10 +4,27 @@ import (
 	"gonum.org/v1/gonum/mat"
 )
 
+// LinkTypeOption gives an enumerated type for the various link types available for rules
+type LinkTypeOption int
+
+const (
```
NGovani, line 10: An explanation of what these mean would be nice

NGovani (changes requested):

internal/common/rules/basicruleevaluator_test.go
Neelesh99: BasicRuleEvaluator feeds off whatever is in the AvailableRulesCache, so this rule is a default entry in there, it was a convenient one to use ;) Defined on from line 16 on globalruleregistration.go

internal/common/rules/rulematrix.go
Neelesh99: Sure added in latest commit

NGovani (approved):

NGovani:
I'm a bit concerned that isn't used yet, so we don't know how its gonna affect end to end test. But I assume that's coming in a future PR it should be fine to merge

Issue #160, opened 2020-12-28
BaseClient functions do not have BASE|OPTIONAL|COMPULSORY note


lhl2617:
As title.

mpardalos:
Can you give an example of this? I've seen these notes in most (if not all) of `baseClient`'s implementations of `Client` methods.

lhl2617:
Mostly iigo

preetl:
Resolved in PR #250

PR #250, opened 2021-01-03
[INFRA] Adding IIGO Baseclient Documentation

preetl:
# Summary

Added documentation to IIGO related baseclient functions to help with agent implementation. I realise that some of the names we have may be confusing so please do let me/team 3/team 4 know if we can help make it clearer.

Re: Issue #160 

## Additional Information

N/A

## Test Plan

N/A


QFSW:

internal/common/baseclient/iigo.go
```diff
@@ -29,36 +31,47 @@ func (c *BaseClient) RuleProposal() string {
...
 }
 
 // GetClientSpeakerPointer is called by IIGO to get the client's implementation of the Speaker Role
+// COMPULSORY: ovverride to return a pointer to your own Speaker object
 func (c *BaseClient) GetClientSpeakerPointer() roles.Speaker {
 	return &BaseSpeaker{}
 }
 
+// BASE
 func (c *BaseClient) TaxTaken(shared.Resources) {
```
NGovani: Shouldn't this be optional as it informs the agent tax has actually been taken from their resources

NGovani (changes requested):
Just one issue otherwise LGTM

internal/common/baseclient/iigo.go
```diff
@@ -29,36 +31,47 @@ func (c *BaseClient) RuleProposal() string {
...
 	return &BaseJudge{}
 }
 
 // GetClientSpeakerPointer is called by IIGO to get the client's implementation of the Speaker Role
+// COMPULSORY: ovverride to return a pointer to your own Speaker object
 func (c *BaseClient) GetClientSpeakerPointer() roles.Speaker {
 	return &BaseSpeaker{}
 }
 
+// BASE
```
lhl2617: what does this func actually do? document it? I think @mpardalos wrote this actually...

preetl: I was just wondering this... I'm inclined to remove it as there is no case that tax wouldn't be taken exactly as expected

internal/common/baseclient/iigo.go
preetl: Do we need this function?

preetl: i've removed the comment for now and made an issue

internal/common/baseclient/iigo.go
preetl: i've made an issue

lhl2617: #261

internal/common/baseclient/iigo.go
NGovani: Yes this function just tell the client that the exchange of resources has actually happened. Since the client may try and give a value that is more than what they currently have which causes an error and means this function isn't called. The function is optional however, and agents don't need to use if they do not want to

NGovani (approved):

PR #254, opened 2021-01-03
Patch: fix #252

### LINKED ISSUE #252: Foraging DeerHunt. Hunting negative deers.
Hi!

I was messing about with the configs such that the foragingMaxDeerPerHunt == foragingDeerMaxPopulation (if you wish to reproduce this error). I was curious what would happen. 

It seems like agents are allowed to hunt negative deers? 

image
image

JamesTev:
# Summary

Fixed #252 where deer returns where always zero when max deer per hunt was set to max population. Note that this should not usually be the case per the design, but this fix ensures it doesn't break the forage in any case. 

## Test Plan

- Tested with max pop and max per hunt equal. Works as expected.


QFSW:

JamesTev:
@E-Mac27 @mpardalos Please have a quick check when you get a moment.

internal/common/foraging/deerhunt.go
```diff
@@ -66,6 +66,10 @@ func (d DeerHunt) getPopulationLinkedProbability(dhConf config.DeerHuntConfig, p
 	thetaCritical := dhConf.ThetaCritical
 	thetaMax := dhConf.ThetaMax
 	p := float64(population) / float64(dhConf.MaxDeerPerHunt)
+
+	if pMax == pCritical { // catch the special case where max deer per hunt = max population. You've got params wrong if you're here though :)
```
E-Mac27, line 70: Haha love the comment

lhl2617, line 70: let's log this so they know :D

E-Mac27 (approved):

internal/common/foraging/deerhunt.go
lhl2617: (you can pass in the server's logger func into `DeerHunt` and make a logger function for DeerHunt)

The output would ideally be

```
[SERVER] [DEERHUNT] ...
```

E-Mac27: I was thinking the same. Just spam the logs but it also might be too much spam? tbf, its already spammy af so feel free to add more.

params.go
```diff
@@ -51,7 +51,7 @@ var (
 	foragingDeerMaxPerHunt = flag.Uint(
 		"foragingMaxDeerPerHunt",
 		4,
-		"Max possible number of deer on a single hunt (regardless of number of participants).",
+		"Max possible number of deer on a single hunt (regardless of number of participants). ** should be strictly less than max deer population.",
```
lhl2617, line 54: you should add a check in the `parseConfig` function which now allows you to return parse errors.

lhl2617 (changes requested):
still impressed with foraging

internal/common/foraging/deerhunt.go
lhl2617: spammy is fine as long as it is _easily filterable_ ( `| grep [DEERHUNT]`)

params.go
lhl2617: if you have this, you can remove your above check (i'd still leave it in)

internal/common/foraging/deerhunt.go
JamesTev: Okay sure - makes sense!

lhl2617: feel free to do this in a future PR--i'm happy with this as long as you add the extra check in `parseConfig`.

params.go
JamesTev: Good idea - will do soon

internal/common/foraging/deerhunt.go
```diff
@@ -81,5 +86,5 @@ func (d DeerHunt) getPopulationLinkedProbability(dhConf config.DeerHuntConfig, p
 
 // Logf is a this type's custom logger
 func (d DeerHunt) Logf(format string, a ...interface{}) {
-	log.Printf("[SERVER][DeerHunt]: %v", fmt.Sprintf(format, a...))
+	log.Printf("[SERVER][DEERHUNT]: %v", fmt.Sprintf(format, a...))
```
lhl2617, line 89: ah, this is fine, but i was actually thinking that when the server first initialises deerhunt, it passes the pointer to the server's logf function (s.Logf). d.Logf would then call the logger with just `[DEERHUNT]` preprended.

lhl2617: so `NewDeerHunt(s.Logf, ...)`

```
type DeerHunt struct {
     logger func(format string, ...)
}
func NewDeerHunt(logger, ...) {
    return DeerHunt{ logger: logger, ... }
}
func (d *DeerHunt) Logf(format, ...) {
    d.logger.Logf("[DEERHUNT]: ...")
}
```

lhl2617: separate PR encouraged :) you can also show how it should be done for  iigointernal, I think they're struggling....

JamesTev: Okay cool. Will do in a separate PR. Let me know if anything else to add in this one pls :) @lhl2617

lhl2617: it's good let's goooooo

lhl2617 (approved):

PR #255, opened 2021-01-03
Minor feat: use server's logger in DeerHunt

JamesTev:
# Summary

Changed DeerHunt's `Logf` implementation to extend the native server `logf` method. 

## Test Plan

- Manually verified using dummy logs.

### DIFF
--- a/internal/common/foraging/setup.go
+++ b/internal/common/foraging/setup.go
@@ -7,12 +7,12 @@ import (
 )
 
 // CreateDeerHunt receives hunt participants and their contributions and returns a DeerHunt
-func CreateDeerHunt(teamResourceInputs map[shared.ClientID]shared.Resources, dhConf config.DeerHuntConfig) (DeerHunt, error) {
+func CreateDeerHunt(teamResourceInputs map[shared.ClientID]shared.Resources, dhConf config.DeerHuntConfig, logger shared.Logger) (DeerHunt, error) {
 	if len(teamResourceInputs) == 0 {
 		return DeerHunt{}, errors.Errorf("No deer hunt resource contributions specified!")
 	}
 	params := deerHuntParams{p: dhConf.BernoulliProb, lam: dhConf.ExponentialRate}
-	return DeerHunt{ParticipantContributions: teamResourceInputs, params: params}, nil // returning error too for future use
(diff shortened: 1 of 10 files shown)


QFSW:

internal/common/foraging/setup.go
```diff
@@ -7,12 +7,12 @@ import (
...
 
 // CreateDeerHunt receives hunt participants and their contributions and returns a DeerHunt
-func CreateDeerHunt(teamResourceInputs map[shared.ClientID]shared.Resources, dhConf config.DeerHuntConfig) (DeerHunt, error) {
+func CreateDeerHunt(teamResourceInputs map[shared.ClientID]shared.Resources, dhConf config.DeerHuntConfig, logger shared.Logger) (DeerHunt, error) {
 	if len(teamResourceInputs) == 0 {
 		return DeerHunt{}, errors.Errorf("No deer hunt resource contributions specified!")
 	}
 	params := deerHuntParams{p: dhConf.BernoulliProb, lam: dhConf.ExponentialRate}
-	return DeerHunt{ParticipantContributions: teamResourceInputs, params: params}, nil // returning error too for future use
+	return DeerHunt{ParticipantContributions: teamResourceInputs, params: params, logger: logger}, nil // returning error too for future use
```
lhl2617, line 15: No problem but note that this causes a runtime error if you try to log within the test. Usually we make a mock logger to save logs in a string so that we can _test logs_--a neat test method

lhl2617: Don't need to change though, it's just a test

lhl2617 (approved):
LGTM!

lhl2617:
We should do this for fish hunt too? (If there's a fish hunt struct)

internal/common/foraging/setup.go
JamesTev: Cool - thanks for the heads up

JamesTev:
> We should do this for fish hunt too? (If there's a fish hunt struct)

Added same functionality for fishing and deer pop. Feel free to merge when you've had a last check.

JamesTev:
> We should do this for fish hunt too? (If there's a fish hunt struct)

Done. Also added for deer population. Feel free to merge when you're happy!

Issue #262, opened 2021-01-04
[IIGO] Move Judge constants to params file


Neelesh99:
Was meant to move earlier, 3 constants at top of judiciary.go should be settable in params file.

Neelesh99:
Fixed with #264

PR #264, opened 2021-01-04
[INFRA/IIGO] Moving constants at top of judiciary to params

Neelesh99:
# Summary

Addressing Issue #262, moving constants defined in judiciary to params file.

## Test Plan

Made appropriate adjustments to tests

### DIFF
--- a/internal/server/iigointernal/judiciary_test.go
+++ b/internal/server/iigointernal/judiciary_test.go
@@ -416,6 +416,7 @@ func TestApplySanctions(t *testing.T) {
 		sanctionRecord     map[shared.ClientID]roles.IIGOSanctionScore
 		sanctionThresholds map[roles.IIGOSanctionTier]roles.IIGOSanctionScore
 		expectedSanctions  []roles.Sanction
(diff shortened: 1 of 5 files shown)


QFSW:

internal/server/iigointernal/judiciary_test.go
```diff
@@ -427,7 +427,7 @@ func TestApplySanctions(t *testing.T) {
 				{
 					ClientID:     shared.Team1,
 					SanctionTier: roles.SanctionTier1,
-					TurnsLeft:    sanctionLength,
+					TurnsLeft:    2,
```
lhl2617, line 432: tests shouldn't break when the defaults in params are changed. you should be injecting these params into your functions.

Neelesh99: Fair enough, I think the functions are injected with these params, the test was poorly written. Refactored to use injected values.

NGovani (approved):
It looks like you've fixed LH's issues. If this passes tests feel free to merge it in

PR #267, opened 2021-01-04
[INFRA/IIGO] Adding IIGO Rules for Monitoring Role Performance

preetl:
# Summary

Added in additional rules to help guide President, Speaker and Judge roles in making decisions in IIGO. Monitoring is now also passed into the judicial, executive and legislative branch. Also amended some return types that were changed by PR #208 .

The baseclient documentation changes are from PR #250 (sorry about that!).

## Additional Information

N/A

## Test Plan

All tests still pass.


QFSW:

internal/server/iigointernal/legislature.go
```diff
@@ -109,35 +111,39 @@ func (l *legislature) RunVote(ruleID string, clientIDs []shared.ClientID) voting
...
-	//TODO: log of vote occurring with ruleID, clientIDs
-	//TODO: log of clientIDs vs islandsAllowedToVote
-	//TODO: log of ruleID vs s.RuleToVote
+
+	variablesToCache := []rules.VariableFieldName{rules.IslandsAllowedToVote}
+	valuesToCache := [][]float64{{float64(len(clientIDs))}}
+	l.monitoring.addToCache(l.SpeakerID, variablesToCache, valuesToCache)
+
+	variablesToCache = []rules.VariableFieldName{rules.SpeakerRuleProposal, rules.PresidentRuleProposal}
+	valuesToCache = [][]float64{{float64(len(ruleID))}, {float64(len(l.ruleToVote))}}
```
E-Mac27: I'm so confused here. You are passing the length of the string ruleID and l.ruleToVote. What happens if there are two different strings of ruleID that have the same length value?

E-Mac27: Is it possible to create some tests so I can understand what's happening?

E-Mac27 (changes requested):

internal/server/iigointernal/legislature.go
preetl: oh shit, yes this is a mistake. ruleid and l.ruletovote are of type string so i think this is a combination of my ide suggesting a change and me blindly following it. I will redo this rule. Doesn't make a lot of sense, sorry!

E-Mac27: Ah okay. No probs.

preetl: Basically i just need to store something so that later the judge can check if they were equal so i've changed it to store a bool now. Should be a touch more logical now!

internal/server/iigointernal/orchestration.go
```diff
@@ -214,6 +219,9 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
 		return false, "President was not apointed by the Judge. Insufficient budget"
 	}
 
+	legislativeBranch.reset()
+	executiveBranch.reset()
```
lhl2617: i very much don't like the reset pattern, why not create a new struct the next time with the right values?

internal/server/iigointernal/orchestration.go
```diff
@@ -70,6 +70,10 @@ func RunIIGO(g *gamestate.GameState, clientMap *map[shared.ClientID]baseclient.C
 		judgeID:           g.JudgeID,
 		internalIIGOCache: []shared.Accountability{},
 	}
+	executiveBranch.monitoring = &monitoring
+	legislativeBranch.monitoring = &monitoring
+	judicialBranch.monitoring = &monitoring
```
lhl2617, line 69: do you mutate these in the branches

lhl2617: why pointers?

preetl: I'm passing in references to the same monitoring object so that they can populate the internal iigo cache variable in monitoring. Is this correct?

internal/server/iigointernal/orchestration.go
preetl: sure thing!

internal/server/iigointernal/orchestration.go
lhl2617: alright, a bit dangerous but ok ._.

E-Mac27 (approved):

PR #269, opened 2021-01-04
[IIGO] Prep for tax/allocation rule evaluation on client side [1/2]

akowal3:
# Summary
- Added transmission of rules & variables for tax/allocation
- Added communication when no decision is made regarding tax/allocation
- Introduced Tax/AllocationDecision struct to wrap all the necessary data into a single struct when sending it to clients

## Auxilary Info
The vast majority of additions is in tests.
Evaluation of rules obtained by the clients in #274.

## Test Plan
Added tax/allocation tests. All tests are passing. IIGO runs


QFSW:

internal/common/baseclient/basepresident.go
```diff
@@ -58,7 +58,8 @@ func (p *BasePresident) SetTaxationAmount(islandsResources map[shared.ClientID]s
 
 	for clientID, clientReport := range islandsResources {
 		if clientReport.Reported {
-			taxAmountMap[clientID] = shared.Resources(float64(clientReport.ReportedAmount) * rand.Float64())
+			taxRate := 0.1
```
lhl2617, line 61: is this a constant?

lhl2617: can it be changed?

lhl2617: ah nvm, just relaised this is baseclient.

akowal3: that is for basepresident, which will be overwritten by client implementation. Because basepresident just shows what the functions can do, it might as well be a constant. would you prefer it to be a config parameter?

akowal3:
will refactor with changes made in #276

PR #270, opened 2021-01-04
Deterministic disaster option

JamesTev:
# Summary

As per discussion with @yp717 @Eirikalb @NGovani. Provides the option to simulate disasters with stochastic _or_ **deterministic** period (guaranteed to occur after _n_ turns). Also renamed `GlobalProb` param in `DisasterConfig` to `Period`. In stochastic case, instead of specifying probability that disaster will occur at each turn, we now specify expected period between disasters. 

## Additional Information

See README.md in disaster package.

## Test Plan

- Added test to verify that deterministic option produces disasters every n turns.


QFSW:

internal/common/config/config.go
```diff
@@ -75,11 +75,12 @@ type FishingConfig struct {
 // DisasterConfig captures disaster-specific config
 type DisasterConfig struct {
 	XMin, XMax, YMin, YMax      shared.Coordinate     // [min, max] x,y bounds of archipelago (bounds for possible disaster)
-	GlobalProb                  float64               // Bernoulli 'p' param. Chance of a disaster occurring
+	Period                      int                   // Period T between disasters in deterministic case and E[T] in stochastic case.
```
lhl2617: uint?

internal/common/disasters/environmentdefinition.go
```diff
@@ -44,24 +44,35 @@ type DisasterEffects struct {
...
+		// E[T] = T (stochastic and deterministic cases respectively). Since
+		// T is a geometric RV in the stochastic case, p = 1/E[T]
+		p := 1 / float64(dConf.Period)
+		pdfGlobal := distuv.Bernoulli{P: p} // Bernoulli RV where `P` = P(X=1)
+
+		if pdfGlobal.Rand() == 1.0 { // D Day
+			dR = DisasterReport{Magnitude: pdfMag.Rand(), X: pdfX.Rand(), Y: pdfY.Rand()}
+		}
+	} else {
+		if turn%uint(dConf.Period) == 0 && turn > 0 {
```
lhl2617, line 71: then cast not needed here if Period is unit

internal/common/config/config.go
JamesTev: dammmit u ryt

NGovani (approved):
LGTM,  if it passes tests go ahead and merge it in

PR #275, opened 2021-01-04
[INFRA/RULES] Infra rules baseclient tools added and tested [1/2]

Neelesh99:
# Summary

As per #265 this PR contains the tools required to give the base clients the required tools for insightful rule evaluation. I've also refactored the BasicEvaluator to be more flexible with regards to which caches it's looking it. For the functionality of the old BasicRuleEvaluator, please use rules.EvaluateRule.

## Additional Information

The next PR will build on these tools with a base client implementation of some level of complexity of ensuring compliance. 

## Test Plan

Refactored old tests, added new tests for new functions.

PR #276, opened 2021-01-04
[INFRA/RULES] Infra rules baseclient functions added

Neelesh99:
# Summary

Added a couple of functions to the basclient that make use of the newly developed toolkit in #275. These include compliance checkers for Tax, Sanctions and Allocations as well as Compliant Value recommendation. 

## Additional Information

Once the new (and exhaustive) set of rules are added, I can extend these baseclient functions to cover them, for now this might provide teams with a  starting point. I've also added a LocalVariableCache to baseclient which stores the required variables that are associated with rules.

## Future tools
Since we don't have feature freeze till the 6th, feel free to write on this PR or PM me what extra functionality you'd like added to baseclient since I recognise this is not all that people wanted, just what I thought might be useful off the top of my head.


QFSW:

Neelesh99:
The high LOC is because this PR is linearly dependant on #275, once that goes in the PR should be like 150 - 200 LOC

internal/common/baseclient/iigo.go
```diff
@@ -17,8 +17,10 @@ func (c *BaseClient) CommonPoolResourceRequest() shared.Resources {
 // the President to help work out how many resources to allocate each island.
 // OPTIONAL : as is, this function will always report island resources accurately
 func (c *BaseClient) ResourceReport() shared.ResourcesReport {
+	amountReported := c.ServerReadHandle.GetGameState().ClientInfo.Resources
+	c.LocalVariableCache[rules.IslandReportedResources] = rules.MakeVariableValuePair(rules.IslandReportedResources, []float64{float64(amountReported)})
 	return shared.ResourcesReport{
-		ReportedAmount: c.ServerReadHandle.GetGameState().ClientInfo.Resources,
+		ReportedAmount: amountReported,
```
Eirikalb, line 23: Nice refactor

internal/common/baseclient/iigo.go
```diff
@@ -56,14 +58,46 @@ func (c *BaseClient) GetClientSpeakerPointer() roles.Speaker {
...
+	isCompliant := c.CheckCompliance(rules.IslandTaxContribution)
+	if isCompliant {
+		// TODO: with this compliance check, agents can see whether they'd like to continue returning this value
+		valToBeReturned = 0
+	}
+	// Use the toolkit to recommend a value
+	newVal, success := c.GetRecommendation(rules.IslandTaxContribution)
+	if success {
+		// TODO: Choose whether to use this compliant value
+		valToBeReturned = shared.Resources(newVal.Values[0])
```
Eirikalb: What is nevVal.Values[0]? could the 0 be encoded as an enum maybe for clarity?

Eirikalb: Same on line 84 and 98

internal/common/rules/clienttools_test.go
```diff
@@ -0,0 +1,505 @@
+package rules
```
Eirikalb, line 1: LOVE THIS FILE! We love tests!

internal/common/rules/clienttools.go
```diff
@@ -0,0 +1,174 @@
+package rules
```
Eirikalb, line 1: These are for the clients to use right? Great addition!

internal/clients/team3/client.go
```diff
@@ -18,9 +18,9 @@ type client struct {
 }
```
Eirikalb, line 18: I suppose this is due to broken backwards compatibility?

Eirikalb (approved):
LGTM, I will pre-approve, but you need to fix linting errors and would be nice to fix that nevVal.Values(0) thing before merge. I dont get github notifications. Overall stuff looks reasonable, thanks for adding to the BaseClient, makes me very happy <3

internal/common/baseclient/iigo.go
Neelesh99: Yeah good point, will do

internal/clients/team3/client.go
Neelesh99: Yup, function name change

internal/common/baseclient/iigo.go
Neelesh99: Done

internal/common/rules/clienttools.go
Neelesh99: Yup

Eirikalb:
Did you fix the linting errors? The build failed....

PR #278, opened 2021-01-04
[IIGO] Outgoing rules proposal

MikeIsMike:
# Summary

Please provide a summary of the change.

## Additional Information

## Test Plan


QFSW:

MikeIsMike:
Updated the whole rule proposal process to use ruleMatrix instead of ruleName. So that the process can handle when an agent wants to propose to change the content of an existing rule.

NGovani (approved):
LGTM if it passes tests then go ahead an merge in.

PR #280, opened 2021-01-05
[INFRA/RULES] Added Link pulling for pulling into and out of caches.

Neelesh99:
# Summary

Quick change, to ensure than when a Linked rule is elected into/out of the RulesInPlay that the rule's Link is also pulled into/ out of cache.

## Test Plan

- Test added

### DIFF
--- a/internal/common/rules/globalrulescache.go
+++ b/internal/common/rules/globalrulescache.go
@@ -38,6 +38,10 @@ func pullRuleIntoPlayInternal(rulename string, allRules map[string]RuleMatrix, p
 		if _, ok := playRules[rulename]; ok {
 			return &RuleError{err: errors.Errorf("Rule '%v' is already in play", rulename), errorType: RuleIsAlreadyInPlay}
 		}
+		linkRule, linked := checkLinking(rulename, allRules)
+		if linked {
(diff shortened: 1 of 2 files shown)


QFSW:

internal/common/rules/globalrulescache.go
```diff
@@ -53,6 +57,10 @@ func PullRuleOutOfPlay(rulename string) error {
 func pullRuleOutOfPlayInternal(rulename string, allRules map[string]RuleMatrix, playRules map[string]RuleMatrix) error {
 	if _, ok := allRules[rulename]; ok {
 		if _, ok := playRules[rulename]; ok {
+			linkRule, linked := checkLinking(rulename, allRules)
+			if linked {
+				delete(playRules, linkRule)
```
NGovani, line 62: Is it possible for linkRule to have another linked rule which would have to either be added to play or removed?

NGovani (approved):
One minor comment but otherwise LGTM

internal/common/rules/globalrulescache.go
Neelesh99: I think from a design perspective this isn't possible, and I wasn't sure whether to recursively look and fail or just ignore like this.

PR #266, opened 2021-01-04
[INFRA/CORE] Make certain config (selectively) visible to client (#263)

lhl2617:
# Summary

#263 . 
- Always visible items are put into `ClientGameConfig`. 
- Selectively ones are put into it under wrapped types such as `SelectivelyVisibleResources`.
- Flags added to make selectively visible values (in)visible

## Test Plan

- Unit tests added
- Changed `func (c *BaseClient) StartOfTurn() {}` to
```
func (c *BaseClient) StartOfTurn() {
	c.Logf("%#v", c.ServerReadHandle.GetGameConfig())
	panic("STOP")
}
```
output in `go run .`
```
...
2021/01/04 14:35:10 [Team6]: config.ClientConfig{CostOfLiving:0, MinimumResourceThreshold:0, MaxCriticalConsecutiveTurns:0x0, DisasterConfig:config.ClientDisasterConfig{MagnitudeResourceMultiplier:config.SelectivelyVisibleFloat64{Value:0, Valid:false}, CommonpoolThreshold:config.SelectivelyVisibleResources{Value:0, Valid:false}}}
...
```
output in `go run . -disasterCommonpoolThresholdVisible true`
```
...
2021/01/04 14:37:36 [Team6]: config.ClientConfig{CostOfLiving:10, MinimumResourceThreshold:5, MaxCriticalConsecutiveTurns:0x3, DisasterConfig:config.ClientDisasterConfig{MagnitudeResourceMultiplier:config.SelectivelyVisibleFloat64{Value:0, Valid:false}, CommonpoolThreshold:config.SelectivelyVisibleResources{Value:50, Valid:true}}}
...
```

NB these tests ran at `85b551d` but behaviour is same for subsequent changes.

PR #284, opened 2021-01-05
[INFRA/IIGO] Allowing clients to access IIGO parameters

Neelesh99:
# Summary

Follow on from LH's PR #266, Clients now have access to all parameters for IIGO.

## Test Plan

- Adjusted tests for new expectations


QFSW:

internal/common/config/clientconfig.go
```diff
@@ -34,3 +62,88 @@ func (c DisasterConfig) GetClientDisasterConfig() ClientDisasterConfig {
 		),
 	}
 }
+
+// GetClientIIGOConfig gets ClientIIGOConfig
+func (c IIGOConfig) GetClientIIGOConfig() ClientIIGOConfig {
+	return ClientIIGOConfig{
+		// Executive branch
+		GetRuleForSpeakerActionCost: getSelectivelyVisibleResources(
```
NGovani: You dont need to call a function just use 
`GetRuleForSpeakerActionCost: c.GetRuleForSpeakerActionCost`

Same for everything below, unless you do intend to to replace `true` with another parameter

internal/common/config/clientconfig.go
```diff
@@ -8,6 +8,33 @@ type ClientConfig struct {
 	MinimumResourceThreshold    shared.Resources
 	MaxCriticalConsecutiveTurns uint
 	DisasterConfig              ClientDisasterConfig
+	IIGOClientConfig            ClientIIGOConfig
+}
+
+// ClientIIGOConfig contains iigo config fields that is visible to clients
+type ClientIIGOConfig struct {
+	// Executive branch
+	GetRuleForSpeakerActionCost        SelectivelyVisibleResources
```
NGovani: just use shared.Resource or int no need to use the selectively visible type

NGovani (changes requested):

internal/common/config/clientconfig.go
Neelesh99: Ohhh, but then the type isn't SelectiveVisibility like the other parameters that are passed. Is that ok?

internal/common/config/clientconfig.go
Neelesh99: Ah right ok, that saves a lot of lines.

internal/common/config/clientconfig.go
Neelesh99: Ok done I get you now

internal/common/config/clientconfig.go
```diff
@@ -8,6 +8,33 @@ type ClientConfig struct {
...
+	RequestAllocationRequestActionCost shared.Resources
+	RequestRuleProposalActionCost      shared.Resources
+	AppointNextSpeakerActionCost       shared.Resources
+	// Judiciary branch
+	InspectHistoryActionCost       shared.Resources
+	InspectBallotActionCost        shared.Resources
+	InspectAllocationActionCost    shared.Resources
+	AppointNextPresidentActionCost shared.Resources
+	SanctionCacheDepth             int
+	HistoryCacheDepth              int
```
lhl2617: these should be `uints`, but ok

internal/common/config/clientconfig.go
```diff
@@ -34,3 +62,31 @@ func (c DisasterConfig) GetClientDisasterConfig() ClientDisasterConfig {
 		),
 	}
 }
+
+// GetClientIIGOConfig gets ClientIIGOConfig
+func (c IIGOConfig) GetClientIIGOConfig() ClientIIGOConfig {
+	return ClientIIGOConfig{
```
lhl2617: since the entire iigo config is exposed linter is telling you to just return `c`

internal/common/config/helpertypes.go
```diff
@@ -35,3 +35,20 @@ func getSelectivelyVisibleResources(value shared.Resources, valid bool) Selectiv
 		Valid: valid,
 	}
 }
+
+// SelectivelyVisibleInteger represents a wrapped int whose value is valid only if the Valid flag is set to true
```
lhl2617: i prefer `SelectivelyVisibleInt` but up to you :)

internal/common/config/clientconfig.go
Neelesh99: Can I do that?

internal/common/config/clientconfig.go
Neelesh99: Fair enough changed

internal/common/config/clientconfig.go
Neelesh99: Screw it, I did that

internal/common/config/helpertypes.go
Neelesh99: Yeah that sounds better tbh

Neelesh99:
Oh shit I fixed the github thing, my commits are now mine :)

NGovani (approved):

PR #287, opened 2021-01-05
[INFRA/CORE] Feature add disasterperiod to clientconf

Eirikalb:
# Summary

Add disasterPeriod to client config as a selectively available value.

## Additional Information

As discussed with @JamesTev earlier today
## Test Plan

Added one testcase which seems to pass 👍


QFSW:

internal/common/config/clientconfig.go
```diff
@@ -13,6 +13,8 @@ type ClientConfig struct {
 // ClientDisasterConfig contains disaster config information visible to clients.
 type ClientDisasterConfig struct {
 	CommonpoolThreshold SelectivelyVisibleResources
+	DisasterPeriod      SelectivelyVisibleDisasterPeriod
+	StochasticDisasters bool // if true, period between disasters becomes random. If false, it will be consistent (deterministic)
```
JamesTev: Why is this not selectively visible too?

JamesTev (approved):
Looks good, thanks @Eirikalb. Haven't looked through this selective visibility config code in detail but back you

internal/common/config/helpertypes.go
```diff
@@ -25,6 +25,16 @@ type SelectivelyVisibleResources struct {
 	Valid bool
 }
 
+type SelectivelyVisibleDisasterPeriod struct {
```
lhl2617: these should be `SelectivelyVisibleUint`

internal/common/config/helpertypes.go
```diff
@@ -25,6 +25,16 @@ type SelectivelyVisibleResources struct {
...
+	Period uint
+	Valid  bool
+}
+
+type SelectivelyVisibleStochasticDisaster struct {
```
lhl2617: `SelectivelyVisibleBool`, similarly.

params.go
```diff
@@ -169,6 +169,16 @@ var (
...
+	disasterPeriodVisible = flag.Bool(
+		"disasterPeriodVisible",
+		true,
+		"Whether disasterPeriod is visible to agents",
+	)
+	disasterStochasticPeriodVisible = flag.Bool(
+		"disasterStochasticPeriodVisible",
+		true,
+		"Whether stochasticPeriod is visible to agents",
+	)
```
lhl2617: I'd move this below the variable where it makes (in)visible. 
Also @JamesTev disasterPeriodStochastic or disasterStochasticPeriod?

JamesTev: Not too fussed. `disasterStochasticPeriod` is how it was defined originally IIRC but the former kind of makes more sense. Up to Eirik.

PR #289, opened 2021-01-05
[INFRA/RULES] Parameterised starting with rules in play

Neelesh99:
# Summary

Start with some rules in play, cos otherwise it's a lawless wasteland. Activated by parameter.


QFSW:

internal/server/server.go
```diff
@@ -79,6 +80,11 @@ func createSOMASServer(
 			CommonPool:  gameConfig.InitialCommonPool,
 		},
 	}
+	if gameConfig.IIGOConfig.StartWithRulesInPlay {
+		for ruleName, _ := range rules.AvailableRules {
```
lhl2617: can you see linter warnings? seems like most people can't because they are not fixed

Neelesh99: XD, as in it doesn't show in IDE, I have to look at PR's to see them. Fixed in latest push

NGovani (approved):
LGTM, not a fan of  Available Rules and Rules In Play being package globals but if it works its all good

PR #290, opened 2021-01-05
[INFRA/VOTING] Patch to Neaten Voting Implementation

preetl:
# Summary

Instead of creating alive client IDs in voting, they are passed in from IIGO.

EDIT: the patch got big

## Additional Information

N/A

## Test Plan

Still pass!

EDIT: added tests!


QFSW:

internal/server/iigointernal/judiciary.go
```diff
@@ -133,7 +133,7 @@ func (j *judiciary) appointNextPresident(monitoring shared.MonitorResult, curren
 			return j.gameState.PresidentID, errors.Errorf("Insufficient Budget in common Pool: appointNextPresident")
 		}
 		election.ProposeElection(shared.President, electionsettings.VotingMethod)
-		election.OpenBallot(electionsettings.IslandsToVote, iigoClients)
+		election.OpenBallot(electionsettings.IslandsToVote, iigoClients, allIslands)
```
lhl2617: just an extra check that only non-dead islands are passed in, right?

preetl: allIslands is passed in from orchestration where it is populated with only alive client IDs

Ning-72 (approved):

NGovani (approved):
LGTM

PR #295, opened 2021-01-06
[INFRA/RULES] Added RulesToPlay to gamestate

Neelesh99:
# Summary

As title.

### DIFF
--- a/internal/common/gamestate/gamestate.go
+++ b/internal/common/gamestate/gamestate.go
@@ -149,9 +149,7 @@ func copyLink(inp rules.RuleLink) rules.RuleLink {
 
 func copyRequiredVariables(inp []rules.VariableFieldName) []rules.VariableFieldName {
 	targetList := make([]rules.VariableFieldName, len(inp))
-	for index, entry := range inp {
-		targetList[index] = entry
-	}
+	copy(targetList, inp)
 	return targetList
 }
 
--- a/internal/common/gamestate/gamestate.go
+++ b/internal/common/gamestate/gamestate.go
@@ -4,7 +4,9 @@ package gamestate
 import (
 	"github.com/SOMAS2020/SOMAS2020/internal/common/disasters"
 	"github.com/SOMAS2020/SOMAS2020/internal/common/foraging"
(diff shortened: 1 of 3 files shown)


QFSW:

internal/common/gamestate/gamestate.go
```diff
@@ -114,6 +120,41 @@ func copyIIGOHistory(iigoHistory map[uint][]shared.Accountability) map[uint][]sh
...
+	return rules.RuleLink{
+		Linked:     inp.Linked,
+		LinkType:   inp.LinkType,
+		LinkedRule: inp.LinkedRule,
+	}
+}
+
+func copyRequiredVariables(inp []rules.VariableFieldName) []rules.VariableFieldName {
+	targetList := make([]rules.VariableFieldName, len(inp))
+	for index, entry := range inp {
```
lhl2617: as per linter :) array copying is built in with `copy(dst, src)`

internal/server/server.go
```diff
@@ -2,14 +2,14 @@ package server
 
 import (
 	"fmt"
+	"github.com/SOMAS2020/SOMAS2020/internal/common/rules"
```
lhl2617: linter

internal/server/server.go
```diff
@@ -59,6 +59,12 @@ func createSOMASServer(
 		forageHistory[t] = make([]foraging.ForagingReport, 0)
 	}
 
+	if gameConfig.IIGOConfig.StartWithRulesInPlay {
+		for ruleName := range rules.AvailableRules {
+			_ = rules.PullRuleIntoPlay(ruleName)
```
lhl2617, line 66: can you add a comment as to why this err is ignored?

lhl2617 (approved):
pre-approving, looks good

internal/server/server.go
Neelesh99: Yup, thanks

lhl2617: I personally still wouldn't ignore btw--at least log it out in case a bug happens so it's easier to debug

PR #300, opened 2021-01-06
[INFRA/IIGO] Removed re-election, added IIGO monitoring cache to gamestate for logging

Spruudis:
Removed re-elections to reduce IIGO complexity. Monitoring cache is now saved across turns so now if a role is caught cheating in the election, (if in play) election rules will oblige the responsible role to hold an election for the no-goodie-election-fraud-island in the next turn (instead of the current). 

Added IIGOCache to the gamestate (updates in monitoring before the cache is cleared)

Changed the order of monitoring so now the president's and judge's monitoring actions are monitored. The speakers monitoring is not (that ok, we said this would be the case from the getgo, the other two are just a happy accident)

### DIFF
--- a/internal/common/gamestate/gamestate.go
+++ b/internal/common/gamestate/gamestate.go
@@ -35,6 +35,9 @@ type GameState struct {
 	// IIGO turns in power (incremented and set by monitoring)
 	IIGOTurnsInPower map[shared.Role]uint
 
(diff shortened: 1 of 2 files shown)


QFSW:

Neelesh99 (approved):
Looks good, thanks for this

Issue #298, opened 2021-01-06
Disasters not taking from the CP


NGovani:
It seems currently disaster do not update/deplete the common pool when applying disaster affects. 
So if there was 100 resources in the CP and a disaster of size 70 struck the islands. The CP would not be reduced to 30 but would stay at 100 after the turn ended

PR #301, opened 2021-01-06
Patch: fix disasters not reducing CP properly

JamesTev:
# Summary

Fix for #298 @NGovani 

## Test Plan

Test isn't the prettiest but seems to serve its purpose. Too tired to do anything fancier.


QFSW:

NGovani:
I added a commit updating the clients resources when depleting their resources.
Initial problem was caused when we changed how resources were taken from the client.  It's causing pull issues. @JamesTev could you give it a quick look I dont have access to a go enviroment rn. Its only 1 line

JamesTev:
> I added a commit updating the clients resources when depleting their resources.
> Initial problem was caused when we changed how resources were taken from the client. It's causing pull issues. @JamesTev could you give it a quick look I dont have access to a go enviroment rn. Its only 1 line

Done

NGovani (approved):
LGTM

PR #309, opened 2021-01-07
Team6 Agent

dejay-vu:
# Summary

Client 6 implementation:
- personality
     - our island can be either **selfish**, **normal**, or **generous** based on how many resources we have
- IITO & friendship
     - gift exchanging is based on friendship with other islands
     - sending and receiving history will be recorded
     - islands will get higher friendship, the more gifts they send us
     - islands having higher friendship with us get more gifts (resources) from us
     - friendship will decrease if we find the island to be selfish
- IIGO
     - will pay the tax if we are not **selfish**
- IIFO
     - the implementation is nearly the same as the baseclient except:
     - predictions will be offered to our good friends
     - predictions are based on past disasters
     - received prediction will be modified by the trust rank (TBD)
- Foraging
     - random foraging at the first 3 round
     - will increase or decrease contribution according to the trend outcome 
     - forage type is based on the contribution.
- voting
     - sorting candidates according to friendship level 
## Additional Information

## Test Plan

- More tests to be added.


QFSW:

olly-larkin (approved):
Lgtm

PR #311, opened 2021-01-07
Team 5 MVP Agent

JamesTev:
# Summary

Team 5 MVP agent, Scooby Doo. Ruh roh RAGGY 🐶 

### Wealth States
These influence our decisions throughout the strategy.
- `Dying`😭, `imperialStudent`😰 , `middleClass` 😒 or `JeffBezos` 🤑 (disclaimer: not linear)

### Opinion Formation
An opinion of another island is our perception of it given past events and is characterised by a score between 1.0 (highest/best perception) and -1.0 (worst/lowest perception). Potential events that may decrease our opinion of other teams:

- allocating/distributing less than they initially offer / we requested for gifting
- corrupt IIGO roles
- nonsensical forecasts
- not sharing information
- being a dick in general

Generally doing the opposite of those things would improve our opinion.

Further, opinions are comprised of more than one `opinionBasis`. We currently store a general `score`, as well as a `forecastingReputation`. We feel that it is necessary to differentiate as we may have a poor general opinion of another team that is stingy with gifting but may simultaneously show skill in forecasting and so would have a high forecasting rep.

#### President 
- We are deciding whether or not we will allocate CP requests depending on the size of the request and the size of the CP
- Set Taxation Amount is proportional to the resources an island has. Else, high flat tax rate is imposed
- We're paying the speaker a salary that is proportional to our wealth level and not always the full amount
- We rig the speaker election based on the opinions that we have on other islands.

#### Speaker
- Used base definition in `BaseSpeaker`

#### Judge
- Used base definition in `BaseJudge`

### IIFO & Forecasting
The forecasting implementation generates forecasts based on several factors.

#### Experience
In predicting the coordinates and magnitude, our simple first approach is to simply average these variables across past disasters. The confidence score associated with these predicted variables takes into account two factors:
- Number of observations (more observations => higher confidence)
- Variance in observations (more variance => lower confidence)

The period is predicted in a similar but slightly different manner. It analyses variance in past disaster periods and looks for a periodic pattern. The degree to which past disasters are periodic influences estimated period confidence (pretty obvious I guess).

The final prediction combines the coordinate and magnitude prediction with that of the period. The combined prediction confidence is computed as a weighted sum of the coordinate+magnitude prediction and the period predition. 

#### Collaboration
The other important component of forecasting is incorporating forecasts from other teams. In each turn, we will receive all forecasts offered to us from other teams (a subset of all teams). A aggregated/collaborative prediction is then formed by computing a *weighted* average based on the following weighting factors for each team:
- Confidence of the team's forecast
- Our current opinion of the team. Specifically, our record of their *forecasting reputation* (explored more below).

### IITO
- Gift at least 1 resource to every agent when we are in the `middleClass`
- Decide gift amount based on our opinions of others, wealth state, our income, and offers
    - no additional gift if
        - you are a dick
        - we have negative income
        - we're `dying`
    - otherwise there will be some

### IIGO 
- Most decisions are based on wealth states. If we are `imperialStudent` (poor as hell), we probably won't give our $ to anyone (including tax)
- Contribute to the common pool based on cashflow of cp
- taxes are always paid unless if we are very poor `imperialStudent`

### Voting
- Vote for roles based on opinion's `score` of other teams 
- The opinion score are ranked/sorted based on the roles history and our general opinion about islands

### Foraging 
- The agent randomly forages accord to how much money he has in the first few turns.
- After that he looks at its history combined with others and picks the one that has the best return.
- If "dying" just gamble it all on deer hunting - life's not worth living.

## Test Plan
- Have added a modest but reasonable number of tests so far. More to come soon.


QFSW:

.github/workflows/golangci-lint.yml
```diff
@@ -2,7 +2,7 @@ name: golangci-lint
 
 on:
   pull_request:
-    branches: [ main ]
+    branches: [ main, develop ]
```
lhl2617, line 5: hmm, yeah that's fine

internal/common/baseclient/forage.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
 )
 
-// DecideForage makes a foraging decision, currently you can only forage deer, but fishing will be available later
+// DecideForage makes a foraging decision
```
lhl2617, line 10: changes should be on a separate non-agent pr

params.go
```diff
@@ -77,7 +77,7 @@ var (
 	)
 	foragingDeerOutputScaler = flag.Float64(
 		"foragingDeerOutputScaler",
-		18,
+		40,
```
lhl2617: default params change? if so should be on separate pr

lhl2617 (changes requested):
as per inline comments

params.go
JamesTev: Yeah apologies - should've picked that up in one of the internal PRs. Reverting.

params.go
```diff
@@ -140,7 +140,7 @@ var (
 	)
 	foragingFishingOutputScaler = flag.Float64(
 		"foragingFishingOutputScaler",
-		12,
+		5,
```
lhl2617: ^

lhl2617 (changes requested):

internal/clients/team5/client.go
```diff
@@ -3,15 +3,149 @@ package team5
...
+//================================================================
+/*	Comunication
+	Gets information on minimum tax amount and cp allocation */
+//=================================================================
+func (c *client) ReceiveCommunication(
+	sender shared.ClientID,
+	data map[shared.CommunicationFieldName]shared.CommunicationContent,
+) {
+	for field, content := range data {
+		switch field {
```
lhl2617: should add default? and log it out at least

go.mod
```diff
@@ -3,10 +3,13 @@ module github.com/SOMAS2020/SOMAS2020
 go 1.15
 
 require (
+	github.com/aclements/go-moremath v0.0.0-20201123183122-222cfcba2589
```
lhl2617, line 6: yarrrrr
image

use at your own risk!

JamesTev: It's been so useful tho :') 😭

lhl2617: this should be fine bc we have `go.sum` to make sure the version matches. I wouldn't worry -- the deadline is in one week and no one's hand is going to be itchy enough to actually update the deps.

JamesTev: Thank you 💕 Will sort out if anything goes wrong ofc.

JamesTev:
Scooby getting sad being in the PR zone for so long now

internal/clients/team5/estimation.go
```diff
@@ -0,0 +1,161 @@
...
+		return pdf, xrange
+	}
+
+	xMax := floats.Max(X)
+	xMin := floats.Min(X)
+
+	if xMax == xMin || stats.Variance(X) == 0 { // special case where bounds are same or zero variance (all same values)
+		return []float64{1.0}, []float64{xMax} // trivial PDF
+	}
+	step := (xMax - xMin) / float64(nSamples)              // 0 check above
```
lhl2617, line 101: div0

internal/clients/team5/estimation.go
```diff
@@ -0,0 +1,161 @@
...
+	pdf, xrange, _ = m.getConstrainedPDF(xMin, xMax, step) // don't
```
lhl2617, line 102: ignoring errors?

internal/clients/team5/forage.go
```diff
@@ -0,0 +1,407 @@
...
+
+// bestForagingType indicates the best foraging method best of RoI (Return on Investment Output/Input - 1)
+func (c *client) bestHistoryForaging(forageHistory forageHistory) shared.ForageType {
+	bestForagingMethod := shared.ForageType(-1) // Default is that there is no good method
+	bestReturn := 0.0
+
+	for forageType, outcomes := range forageHistory { // For each foraging type
+		returnOI := 0.0
+		for _, returns := range outcomes {
+			returnOI += float64((returns.output / returns.input) - 1) // Cumlative sum of the return on investment
```
lhl2617: do a quick search for `/` to check if you've checked div0?

lhl2617 (approved):
go go go whenever you're ready. make sure there are no functional changes in non-client code. please also check for div0

internal/clients/team5/estimation.go
JamesTev: Will never reach here - it returns if nSamples = 0 above

internal/clients/team5/estimation.go
JamesTev: The only cases where `getConstrainedPDF` throws errors are handled above in this method. I could be more explicit if need be though

internal/clients/team5/forage.go
JamesTev: Will do 👍

PR #317, opened 2021-01-07
[VIS] Log logs when runGame fails

lhl2617:
Just for convenience of error reporting and debugging. Related to #310 .


QFSW:

olly-larkin (approved):
image

lgtm

PR #322, opened 2021-01-07
Add Nan and div 0 checks for IIFO

### LINKED ISSUE #310: [VIS] Setting costOfLiving to 5 breaks the simulation
## Steps
1. Set `costOfLiving` to 5
2. Run simulation

## Results
image

## Notes
Resetting flags doesn't fix it, only refresh
`go run . --costOfLiving 5` works fine, so its a website specific error

Eirikalb:
# Summary

Fix #310  nan values in IIFO.

## Additional Information

## Test Plan

### DIFF
--- a/internal/server/iifo.go
+++ b/internal/server/iifo.go
@@ -49,7 +49,7 @@ func (s *SOMASServer) getPredictions() shared.DisasterPredictionInfoDict {
 	for _, id := range nonDeadClients {
 		c := s.clientMap[id]
 		disasterPrediction := c.MakeDisasterPrediction()
-		if math.IsNaN(disasterPrediction.PredictionMade.Confidence) || disasterPrediction.PredictionMade.TimeLeft == -9223372036854775808 || math.IsNaN(disasterPrediction.PredictionMade.Magnitude) || math.IsNaN(disasterPrediction.PredictionMade.CoordinateY) || math.IsNaN(disasterPrediction.PredictionMade.CoordinateX) {
+		if math.IsNaN(disasterPrediction.PredictionMade.Confidence) || math.IsNaN(disasterPrediction.PredictionMade.Magnitude) || math.IsNaN(disasterPrediction.PredictionMade.CoordinateY) || math.IsNaN(disasterPrediction.PredictionMade.CoordinateX) {
 			continue
 		}
 		islandPredictionsDict[id] = disasterPrediction
--- a/internal/server/iifo.go
+++ b/internal/server/iifo.go
@@ -1,6 +1,8 @@
 package server
 
 import (
+	"math"
+
 	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
 )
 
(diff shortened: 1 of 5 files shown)


QFSW:

preetl (approved):
looks good to me 👍🏾

internal/server/iifo.go
```diff
@@ -46,8 +48,13 @@ func (s *SOMASServer) getPredictions() shared.DisasterPredictionInfoDict {
 	nonDeadClients := getNonDeadClientIDs(s.gameState.ClientInfos)
 	for _, id := range nonDeadClients {
 		c := s.clientMap[id]
-		islandPredictionsDict[id] = c.MakeDisasterPrediction()
+		disasterPrediction := c.MakeDisasterPrediction()
+		if math.IsNaN(disasterPrediction.PredictionMade.Confidence) || disasterPrediction.PredictionMade.TimeLeft == -9223372036854775808 || math.IsNaN(disasterPrediction.PredictionMade.Magnitude) || math.IsNaN(disasterPrediction.PredictionMade.CoordinateY) || math.IsNaN(disasterPrediction.PredictionMade.CoordinateX) {
```
lhl2617:
```
disasterPrediction.PredictionMade.TimeLeft == -9223372036854775808
```

this will fail on a 32-bit system

lhl2617: IMO we should not _produce_ `NaN`s/or min value. this should be handled at creation time, but this gets values from client side so here is fine.

lhl2617: see https://stackoverflow.com/questions/6878590/the-maximum-value-for-an-int-type-in-go for solution--otherwise, I'd ignore checking for that int. 

also, i think client code for team 3 also produces `NaN`s atm.

lhl2617 (dismissed):
remove magic num for min 64bit int

internal/server/iifo.go
Eirikalb: Should I make <math.MinINT32 or something then?

lhl2617: I feel like `TimeLeft` should be a `uint`? then we don't really need to do anything about it.

Eirikalb: Made it a uint but it could still be overflowed... 
````
2021/01/07 17:29:51 [Team6]: Final Prediction: [{NaN NaN NaN 9223372036854775808 NaN}]
2021/01/07 17:29:51 [Team1]: Final Prediction: [{1.2 1.2 1.2 1 99}]
2021/01/07 17:29:51 [Team2]: Final Prediction: [{1.2 1.2 1.2 1 99}]
2021/01/07 17:29:51 [Team3]: Final Prediction: [{NaN NaN NaN 9223372036854775808 NaN}]
2021/01/07 17:29:51 [Team4]: Final Prediction: [{1.2 1.2 1.2 1 99}]
2021/01/07 17:29:51 [Team5]: Final Prediction: [{1.2 1.2 1.2 1 99}]
````

Eirikalb: The final prediction is not used anywhere by team3 and team6 yet and is never distributed so its not breaking the codebase AFAIK

Issue #327, opened 2021-01-07
[VIS] panic during cost of living = 5


lhl2617:
![image](https://user-images.githubusercontent.com/33488131/103928184-61161e00-5113-11eb-9ff7-df11c8dd6563.png)

## Repro

https://somas2020.github.io/SOMAS2020/#/newrun

Change `costOfLiving` to 5

Run

![image](https://user-images.githubusercontent.com/33488131/103928233-7428ee00-5113-11eb-8d3d-332e5add8203.png)

console logs

![image](https://user-images.githubusercontent.com/33488131/103928256-7db25600-5113-11eb-9538-fc9452b33cb6.png)

lhl2617:
can be reproduced by @QFSW as well

lhl2617:
panicking line:
https://github.com/SOMAS2020/SOMAS2020/blob/de7dcd4070831255aa2091e7688fc59c17cd66cd/internal/clients/team2/client.go#L157

assigning to @Eirikalb, blame author @Hamish989

lhl2617:
offending commit https://github.com/SOMAS2020/SOMAS2020/commit/07ba2a41d9f014440888a044dcb1bf0b164e221a author @Hamish989

lhl2617:
see #330 for recover and better logs and ui when panic

PR #330, opened 2021-01-07
[VIS/WASM] Try to recover in RunGame if panicked

lhl2617:
# Summary

When the WASM instance panics it just shows stuff like #327. Add a recover and handle the panic.

## Additional Info
- Change prettier settings to account for windows crlf issue

## Test Plan

Try to repo as per #327. Output:

image


Neelesh99 (approved):
LGTM

PR #338, opened 2021-01-07
[VIS] actually fixed horizontal scroll

olly-larkin:
# Summary

LH has solved everything 

## Additional Information

NA

## Test Plan

Different browser sizes


lhl2617 (approved):
maybe

PR #340, opened 2021-01-08
[IITO/VIS] Add iito transactions to gamestate

Eirikalb:
# Summary

Add IITO transactions to gamestate so that visualization can use the JSON output

## Additional Information

## Test Plan

- Visual inspection....


yp717 (approved):
lgtm

PR #342, opened 2021-01-08
[VIS] Resources graph update, about page added

augusta100:
# Summary

Critical threshold level and common pool resources added to resources graph. About page added. 

Screenshot 2021-01-08 at 10 34 58
Screenshot 2021-01-08 at 10 35 12


website/src/components/About/About.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+const About = () => {
+  return (
+    <div style={{ paddingTop: 24 }}>
```
yp717: can you add more padding on the top here? something like paddingTop: 50px I think would work better

website/src/components/Visualisations/Resources/LineGraph.tsx
```diff
@@ -44,15 +47,16 @@ class LineRechartComponent extends React.Component<IProps, any> {
...
         team5: 'ClientInfos.Team5.Resources',
         team6: 'ClientInfos.Team6.Resources',
+        CommonPool: 'CommonPool',
       },
       chartData: this.props.output,
     }
   }
 
-  handleClick = (dataKey) => {
+  handleClick = (dataKey: any) => {
```
Darrekt: why was this necessary?

website/src/components/Visualisations/Resources/LineGraph.tsx
```diff
@@ -44,15 +47,16 @@ class LineRechartComponent extends React.Component<IProps, any> {
...
     this.setState({
       disabled: this.state.disabled.includes(dataKey)
-        ? this.state.disabled.filter((obj) => obj !== dataKey)
+        ? this.state.disabled.filter((obj: any) => obj !== dataKey)
```
Darrekt: same here, in general the use of `any` is frowned upon and can lead to a lot of runtime errors later on.

yp717 (changes requested):
aside from the top spacing lgtm - there are some any types that could maybe be typed a bit more strictly but @Darrekt will address those :)

Darrekt (changes requested):
Looks good but I would like to help you address the `any` bits. Let me know if you want to sit down about it together.

website/src/components/Visualisations/Resources/LineGraph.tsx
augusta100: It was throwing an error and that was the quick fix, changed it to string now, is that okay?

website/src/components/Visualisations/Resources/LineGraph.tsx
augusta100: Same with this one, changed to string

yp717 (approved):
lgtm

website/src/components/Visualisations/Resources/LineGraph.tsx
```diff
@@ -16,26 +16,54 @@ import {
...
   ReferenceLine,
+  ReferenceArea,
+  TooltipProps,
 } from 'recharts'
 import _ from 'lodash'
 import { OutputJSONType } from '../../../consts/types'
+import styles from './Resources.module.css'
+
+const CustomTooltip = (props: TooltipProps) => {
+  const { active, label, payload } = props
```
Darrekt:
```suggestion
const CustomTooltip = ({ active, label, payload }: TooltipProps) => {
```

Darrekt (approved):

PR #346, opened 2021-01-08
Removing Globals Part 2 of 4 (probably)

Neelesh99:
# Summary

The Variable map has been removed from global scope, now encapsulated in the game state under the RuleContext struct. Should help with fixing WASM NaN outputs.

## Additional Information

I've had to adjust one or two client implementations to suit the new non-global maps, pls let me know if I've done it wrong.

## Test Plan

Adjusted a bunch of tests to fit with the no-longer-global VariableMap

P.S. Both us (Team 3) and Team 6 (@Regnigh) didn't instantiate our baseRoles when loading up the custom role implementations. I've instantiated them in this PR, we'll have to be wary of this in the future. (Thanks @lhl2617 for the help finding this issue)


internal/clients/team2/client.go
```diff
@@ -184,7 +184,7 @@ func (c *client) gameConfig() config.ClientConfig {
 // You will need it to access the game state through its GetGameStateMethod.
 func (c *client) Initialise(serverReadHandle baseclient.ServerReadHandle) {
 	c.ServerReadHandle = serverReadHandle
-	c.LocalVariableCache = rules.CopyVariableMap()
+	c.LocalVariableCache = rules.CopyVariableMap(c.ServerReadHandle.GetGameState().RulesInfo.VariableMap)
```
NGovani, line 187: make sure to confirm with team 2 when changing this code

NGovani (approved):
just check with the teams whose code your changing, also linter issues otherwise LGTM.
Pre approving would be good for LH to check it aswell

internal/clients/team2/client.go
Eirikalb: Team 2 approves

Neelesh99: Cheers

PR #354, opened 2021-01-08
Removing Globals part 3 of 4

Neelesh99:
# Summary

Finished removing global variables, all rules information is contained entirely within RuleContext struct in gamestate or in client local caches.

## Test Plan

Had to modify many tests, all now passing.

### DIFF
--- a/internal/common/baseclient/baseSpeaker.go
+++ b/internal/common/baseclient/baseSpeaker.go
@@ -1,16 +1,18 @@
 package baseclient
 
 import (
+	"github.com/SOMAS2020/SOMAS2020/internal/common/gamestate"
 	"github.com/SOMAS2020/SOMAS2020/internal/common/rules"
 	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
 )
 
 type BaseSpeaker struct {
+	gameState gamestate.ClientGameState
 }
 
 // PayJudge is used for paying judge for his service
 func (s *BaseSpeaker) PayJudge() shared.SpeakerReturnContent {
-	JudgeSalaryRule, ok := rules.RulesInPlay["salary_cycle_judge"]
+	JudgeSalaryRule, ok := s.gameState.RulesInfo.CurrentRulesInPlay["salary_cycle_judge"]
 	var JudgeSalary shared.Resources = 0
 	if ok {
 		JudgeSalary = shared.Resources(JudgeSalaryRule.ApplicableMatrix.At(0, 1))
--- a/internal/common/baseclient/baseSpeaker.go
+++ b/internal/common/baseclient/baseSpeaker.go
@@ -7,12 +7,12 @@ import (
 )
 
 type BaseSpeaker struct {
-	gameState gamestate.ClientGameState
+	GameState gamestate.ClientGameState
 }
 
 // PayJudge is used for paying judge for his service
 func (s *BaseSpeaker) PayJudge() shared.SpeakerReturnContent {
-	JudgeSalaryRule, ok := s.gameState.RulesInfo.CurrentRulesInPlay["salary_cycle_judge"]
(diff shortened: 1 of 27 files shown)


internal/server/iigointernal/executive_test.go
```diff
@@ -690,8 +690,10 @@ func TestReplyAllocationRequest(t *testing.T) {
 		},
 	}
 
-	rules.PullRuleIntoPlay("allocation_decision")
-	rules.PullRuleIntoPlay("check_allocation_rule")
+	rulesInPlay := map[string]rules.RuleMatrix{}
+
+	rules.PullRuleIntoPlayInternal("allocation_decision", generateRuleStore(), rulesInPlay)
```
lhl2617: even in a test these should be checked

internal/server/iigointernal/executive_test.go
```diff
@@ -894,8 +910,8 @@ func TestBroadcastTaxation(t *testing.T) {
 		},
 	}
 
-	rules.PullRuleIntoPlay("tax_decision")
-	rules.PullRuleIntoPlay("check_taxation_rule")
+	rules.PullRuleIntoPlayInternal("tax_decision", generateRuleStore(), rulesInPlay)
```
lhl2617: +1

lhl2617:
can you update from main? thx!

internal/common/baseclient/baseSpeaker.go
```diff
@@ -1,16 +1,18 @@
 package baseclient
 
 import (
+	"github.com/SOMAS2020/SOMAS2020/internal/common/gamestate"
 	"github.com/SOMAS2020/SOMAS2020/internal/common/rules"
 	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
 )
 
 type BaseSpeaker struct {
+	gameState gamestate.ClientGameState
```
NGovani: doesnt this need to be initialised in GetClientSpeakerPointer() in baseclient iigo

NGovani (changes requested):
one issue otherwise LGTM

internal/server/iigointernal/executive_test.go
Neelesh99: Oooo nice catch, changed in latest commit

internal/server/iigointernal/executive_test.go
Neelesh99: As above, thanks for finding it

internal/common/baseclient/baseSpeaker.go
Neelesh99: Ah I knew I forgot smth, thanks, changed in latest commit

Neelesh99:
> can you update from main? thx!

Updated

lhl2617:
>  (no, it was never 4) 

what about monitoring? 
image

lhl2617:
game still crashing after this fix.

Neelesh99:
> > (no, it was never 4)
> 
> what about monitoring?
> image

Dammit, you're right. @preetl is working on it. I believe her computer had decided it wasn't going to work on it until about an hour ago.

Neelesh99:
> game still crashing after this fix.

Hard crash in terminal or the NaN thing?

lhl2617:
> > game still crashing after this fix.
> 
> Hard crash in terminal or the NaN thing?

 NaN

Neelesh99:
> > > game still crashing after this fix.
> > 
> > 
> > Hard crash in terminal or the NaN thing?
> 
> NaN

Ok, hopefully monitoring is the culprit then.

lhl2617:
ok. monitoring is not the culprit, but let's remove it as well. the problem is in the defualt president team 3.

Neelesh99:
> ok. monitoring is not the culprit, but let's remove it as well. the problem is in the default president team 3.

Damn that Team 3, always messing with our stuff

NGovani (approved):
Lgtm, if it works in testing merge it in

PR #360, opened 2021-01-08
[INFRA/IIGO] Removing Globals from IIGO : Part 4/4 ... maybe more

preetl:
# Summary

Moved IIGOSanctionCache and IIGOHistoryCache to game state. 
Building up to removing monitoring. Sorry about LOC, had to move some structs because of an import cycle.

## Additional Information

N/A

## Test Plan

Still passing!

PR #361, opened 2021-01-08
[IIGO] (as usual) Caches are reset due to use of syncWithGame

Neelesh99:
# Summary

Removed j.resetCaches from judiciary's syncWithGame.

## Additional Information

syncWithGame wasn't initially intended to be called every turn. It was there to facilitate hot-swapping of game-configs (for some reason I thought that was a thing). Judiciary branch's syncWithGame function also reset its caches, deffo not smth we wanna do each turn.


preetl (approved):
cache me outside

lhl2617:
if resetCaches is never called can we also remove that thing :D

Neelesh99:
> if resetCaches is never called can we also remove that thing :D

Sure, I'll just wait for #360 to merge and pull into this branch.

PR #363, opened 2021-01-08
[INFRA/IIGO] Further remove some globals

lhl2617:
# Summary

- Change global constant maps to be a function returning that map
- Change global constant slices to be a slice literal

## Additional Information

Should be the final one for #125

## Test Plan

In CI we trust


Neelesh99 (approved):
Thanks for these, completely forgot I made this. LGTM

PR #365, opened 2021-01-09
[INFRA/CORE] Add logging level in CLI

lhl2617:
# Summary

Logging slows down running. Add described logging levels possible in CLI flag.

## Additional Info
Add a warning about deletion in `--output` flag.

## Test Plan

Tested with
`go run . --logLevel 0|1|2` -- all behaviour as expected.

### DIFF
--- a/main.go
+++ b/main.go
@@ -31,12 +31,23 @@ var (
 	outputFolderName = flag.String(
 		"output",
 		"output",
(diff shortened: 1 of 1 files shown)


Neelesh99 (approved):
LGTM, this will save so much time.

PR #367, opened 2021-01-09
add IIGOElection to gameState

ramon15749:
# Summary

Add IIGO Election into the gamestate so it is in output.json and can be used to visualise

## Additional Information

Create a new struct VotingInfo in gamestate to contain all the necessary information in voting


internal/server/iigointernal/orchestration.go
```diff
@@ -228,6 +228,8 @@ func RunIIGO(logger shared.Logger, g *gamestate.GameState, clientMap *map[shared
 	var appointJudgeError, appointSpeakerError, appointPresidentError error
 	actionCost := gameConf.IIGOConfig
 	costOfElection := actionCost.AppointNextSpeakerActionCost + actionCost.AppointNextJudgeActionCost + actionCost.AppointNextPresidentActionCost
+	//reset IIGO Election before using it
+	g.IIGOElection = nil
```
Neelesh99: I know obv we use the append function so it takes care of the nil, but could we maybe just instantiate to empty or smth, just in case we want to check this down the line and it becomes nil dereference.

Neelesh99 (approved):
Pre-approving, if you can pls address the comment, it's functionally not broken but a little scary. If someone from Vis could have a look that would be swell as well.

internal/server/iigointernal/orchestration.go
ramon15749: `g.IIGOElection = make([]gamestate.VotingInfo, 0)`

did this instead hopefully thats alright

Neelesh99: Yup, looks good

internal/common/gamestate/gamestate.go
```diff
@@ -233,6 +236,14 @@ type ClientInfo struct {
 	// REMEMBER TO EDIT `Copy` IF YOU ADD ANY REFERENCE TYPES (maps, slices, channels, functions etc.)
```
lhl2617, line 243: Did you edit copy to copy this?

ramon15749: sorry i forgot 😢, its added now

lhl2617 (approved):
lgtm

preetl (approved):
Happy with this!

PR #368, opened 2021-01-09
Emergency NaN patch 3

Neelesh99:
# Summary

Added safeDivResources, to guard against NaN's. Not sure if this algo was actually producing any but for safety sake.

## Additional Information

Much like the fast and furious series, we seem to keep making these and no-one seems to know why or how.


internal/clients/team3/clientiigo.go
```diff
@@ -339,3 +339,10 @@ func copyRulesMap(inp map[string]rules.RuleMatrix) map[string]rules.RuleMatrix {
 	}
 	return newMap
 }
+
+func safeDivResources(numerator shared.Resources, denominator shared.Resources) shared.Resources {
+	if denominator != 0 {
+		return numerator / denominator
+	}
+	return numerator
```
lhl2617, line 347: I'd log when this happens tbh -- btw I like returning 0 instead.

lhl2617 (approved):
image

internal/clients/team3/clientiigo.go
Neelesh99: It's our agent strat to do the illogical thing

PR #370, opened 2021-01-09
[IIGO] Preventing premature elections

Neelesh99:
# Summary

A number of misc. bugs in IIGO conspired to keep electing out certain roles from office too early. We have fixed this behaviour.

## Additional Information

Pls be very careful around your CallRoleElection implementations, it is quite easy to cause IIGO to accidentally descend into anarchy.

### DIFF
--- a/internal/server/iigointernal/monitoring.go
+++ b/internal/server/iigointernal/monitoring.go
@@ -45,7 +45,7 @@ func (m *monitor) monitorRole(roleAccountable baseclient.Client) shared.MonitorR
 		}
 
 		if !evaluationResult {
-			m.gameState.IIGOTurnsInPower[roleName] = m.config.IIGOConfig.IIGOTermLengths[roleName]
+			m.gameState.IIGOTurnsInPower[roleName] = m.config.IIGOConfig.IIGOTermLengths[roleName] + 1
 		}
 
 		m.Logf("Monitoring of %v result %v ", roleToMonitor, evaluationResult)
--- a/internal/server/iigointernal/monitoring.go
+++ b/internal/server/iigointernal/monitoring.go
@@ -59,8 +59,6 @@ func (m *monitor) monitorRole(roleAccountable baseclient.Client) shared.MonitorR
 
 			message := generateMonitoringMessage(roleName, evaluationResult)
 			broadcastToAllIslands(m.iigoClients, roleAccountable.GetID(), message, *m.gameState)
-
-			m.gameState.IIGOTurnsInPower[roleName]++
 		}
 
 		result := shared.MonitorResult{Performed: decideToMonitor, Result: evaluationResult}
@@ -72,10 +70,10 @@ func (m *monitor) monitorRole(roleAccountable baseclient.Client) shared.MonitorR
 
 func (m *monitor) evaluateCache(roleToMonitorID shared.ClientID, ruleStore map[string]rules.RuleMatrix) bool {
 	performedRoleCorrectly := true
+	var rulesAffected []string
 	for _, entry := range m.gameState.IIGORoleMonitoringCache {
 		if entry.ClientID == roleToMonitorID {
 			variablePairs := entry.Pairs
-			var rulesAffected []string
 			for _, variable := range variablePairs {
 				valuesToBeAdded, foundRules := rules.PickUpRulesByVariable(variable.VariableName, ruleStore, m.gameState.RulesInfo.VariableMap)
 				if foundRules {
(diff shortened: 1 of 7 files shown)


internal/server/iigointernal/monitoring.go
```diff
@@ -59,8 +59,6 @@ func (m *monitor) monitorRole(roleAccountable baseclient.Client) shared.MonitorR
 
 			message := generateMonitoringMessage(roleName, evaluationResult)
 			broadcastToAllIslands(m.iigoClients, roleAccountable.GetID(), message, *m.gameState)
-
-			m.gameState.IIGOTurnsInPower[roleName]++
```
Spruudis, line 63: m.gameState.IIGOTurnsInPower[roleName] = config.IIGOTermlengths[roleName] + 1

internal/server/server.go
```diff
@@ -94,9 +95,9 @@ func createSOMASServer(
 				shared.Judge:     0,
 				shared.Speaker:   0,
 			},
-			SpeakerID:   shared.Team1,
-			JudgeID:     shared.Team2,
-			PresidentID: shared.Team3,
+			SpeakerID:   shared.ClientID(rand.Intn(len(shared.TeamIDs))),
+			JudgeID:     shared.ClientID(rand.Intn(len(shared.TeamIDs))),
+			PresidentID: shared.ClientID(rand.Intn(len(shared.TeamIDs))),
```
Spruudis: there is a PR from Toby coming assigning 3 different random islands, just fyi

Spruudis: see #371 :D

Spruudis (changes requested):

internal/server/iigointernal/monitoring.go
Spruudis: or something along those lines. At some point somebody had misinterpreted and changed this to ++

ezgimez: @Neelesh99

preetl: that would be me :D why would this be in if announce {} though ?

internal/server/server.go
Neelesh99: Yeah I had a look at #371, why is it so long, doesn't this method do the same thing in 3 lines?

internal/server/iigointernal/monitoring.go
Neelesh99: We removed the ++ line because it double incremented the TurnsInPower, would the suggested line do more than a double increment?

internal/server/server.go
tobhil98: I spoke to Rudolfs about it and we wanted the three roles to be unique at the start.

internal/server/iigointernal/monitoring.go
Spruudis: it sets it so the responsible role would have rule obliging them to remove the monitored role from power. If the term length is 6 and we are in turn 3, this would set turns in power to 7 (and now an election is permitted and obliged to be held)

Spruudis: as to why, this is what we defined the announce power to be. If there is an announcement of misconduct an election should be held

preetl: so this should happen if the monitoringResult announced is negative right? not if announce ?

internal/server/server.go
ezgimez: I need to speak with @Spruudis

ezgimez: like, I don't remember him saying this to me

ezgimez: @Spruudis  and WHOMEVER needs to be, 

quick call in IIGO server?

internal/server/iigointernal/monitoring.go
Spruudis: yup

internal/server/iigointernal/monitoring.go
```diff
@@ -42,6 +44,10 @@ func (m *monitor) monitorRole(roleAccountable baseclient.Client) shared.MonitorR
 			evaluationResult = m.evaluateCache(roleToMonitor, m.gameState.RulesInfo.CurrentRulesInPlay)
 		}
 
+		if !evaluationResult {
+			m.gameState.IIGOTurnsInPower[roleName] = m.config.IIGOConfig.IIGOTermLengths[roleName]
```
Spruudis:
```suggestion
			m.gameState.IIGOTurnsInPower[roleName] = m.config.IIGOConfig.IIGOTermLengths[roleName] + 1
```
for term to have 'ended' when cheating is found

Neelesh99: Ok cheers, I'll commit once I've cleared a different bug.

internal/server/server.go
Neelesh99: Ok fixed

Spruudis (approved):
👁️👄👁️

internal/server/server.go
```diff
@@ -13,6 +11,7 @@ import (
 	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
 	"github.com/SOMAS2020/SOMAS2020/internal/server/iigointernal"
 	"github.com/pkg/errors"
+	"log"
```
lhl2617: where is my goimport linter

params.go
```diff
@@ -27,7 +27,7 @@ var (
 	)
 	initialCommonPool = flag.Float64(
 		"initialCommonPool",
-		100,
+		1000,
```
lhl2617, line 30: why?

lhl2617: (provide reason in desc)

internal/server/server.go
Neelesh99: Yup, fair enough fixed it

params.go
Neelesh99: IIGO never ran ;( with 100

lhl2617: image

PR #371, opened 2021-01-09
[INFRA/IIGO] Role initialisation

tobhil98:
# Summary

- Randomly distribute roles at the start. 

## Test Plan

- Simple tests to make sure it worked


internal/server/server.go
```diff
@@ -172,3 +174,31 @@ func (s ServerForClient) GetGameState() gamestate.ClientGameState {
 func (s ServerForClient) GetGameConfig() config.ClientConfig {
 	return s.server.gameConfig.GetClientConfig()
 }
+
+func randomAssign(input []shared.ClientID) (shared.ClientID, shared.ClientID, shared.ClientID) {
+	lst := arrayCopy(input)
+
```
lhl2617:
```suggestion
	lst = make([]shared.ClientID, len(input))
	copy (lst, input)
```

internal/server/server.go
```diff
@@ -172,3 +174,31 @@ func (s ServerForClient) GetGameState() gamestate.ClientGameState {
...
+	// Just randomly assign roles if there are not enough clients
+	if len(lst) < 3 {
+		return lst[rand.Intn(len(lst))], lst[rand.Intn(len(lst))], lst[rand.Intn(len(lst))]
+	}
+
+	uniqueClients := make([]shared.ClientID, 0, 3)
+	for index := 0; index != 3; index++ {
```
lhl2617: why not `i` and `i < 3`?

internal/server/server.go
tobhil98: Perfect, thanks! Been trying to find something better

internal/server/server.go
```diff
@@ -172,3 +174,31 @@ func (s ServerForClient) GetGameState() gamestate.ClientGameState {
 func (s ServerForClient) GetGameConfig() config.ClientConfig {
 	return s.server.gameConfig.GetClientConfig()
 }
+
+func randomAssign(input []shared.ClientID) (shared.ClientID, shared.ClientID, shared.ClientID) {
+	lst = make([]shared.ClientID, len(input))
```
lhl2617: don't just commit changes btw--need to remove `arrayCopy`.

tobhil98: I will

internal/server/server.go
tobhil98: Updated

internal/server/server.go
```diff
@@ -172,3 +174,31 @@ func (s ServerForClient) GetGameState() gamestate.ClientGameState {
 func (s ServerForClient) GetGameConfig() config.ClientConfig {
 	return s.server.gameConfig.GetClientConfig()
 }
+
+func randomAssign(input []shared.ClientID) (shared.ClientID, shared.ClientID, shared.ClientID) {
```
lhl2617: nitpick: I'd rename this to `randomAssignIIGORoles`, `randomAssign` is a very generic name.

tobhil98: Done

internal/server/server.go
```diff
@@ -172,3 +174,23 @@ func (s ServerForClient) GetGameState() gamestate.ClientGameState {
 func (s ServerForClient) GetGameConfig() config.ClientConfig {
 	return s.server.gameConfig.GetClientConfig()
 }
+
+func randomUniqueAssignIIGORoles(input []shared.ClientID) (shared.ClientID, shared.ClientID, shared.ClientID) {
+	lst := make([]shared.ClientID, len(input))
+	copy(lst, input)
+	// Just randomly assign roles if there are not enough clients
```
lhl2617, line 181: this is fine to me, but you could do this:

```
// shuffle lst
sort.Shuffle(len(lst), func(i, j int) { lst[i], lst[j] = lst[j], lst[i] }
// if >= 3, return the first three
if len(lst) >= 3 {
    return lst[0], lst[1], lst[2]
}
// randomly pick three
return lst[rand.Intn(len(lst))], lst[rand.Intn(len(lst))], lst[rand.Intn(len(lst))]

File: internal/server/server.go [hunk 6]
```diff
@@ -172,3 +174,23 @@ func (s ServerForClient) GetGameState() gamestate.ClientGameState {
...
+		return lst[rand.Intn(len(lst))], lst[rand.Intn(len(lst))], lst[rand.Intn(len(lst))]
+	}
+
+	uniqueClients := make([]shared.ClientID, 0, 3)
+	for index := 0; index < 3; index++ {
+		randomNum := rand.Intn(len(lst))
+		uniqueClients = append(uniqueClients, lst[randomNum])
+		// Move selected client to the end of the list and remove it
+		lst[randomNum] = lst[len(lst)-1]
+		lst = lst[:len(lst)-1]
```
lhl2617: this isn't correct btw, i think you meant to swap: 

```
lst[len(lst)-1], lst[randomNum] = lst[randomNum], lst[len(lst)-1]
```

but I strongly prefer the above approach. there be dragons when you play with slices like this when not needed.

internal/server/server.go
lhl2617: see: https://play.golang.org/p/fBhw9ZTBK9v

tobhil98: What do you recommend that I do with the error? createSOMASServer does not return an error. Should I just panic? I was hoping that if there were no clients it would get picked up before this stage.

internal/server/server.go
```diff
@@ -172,3 +174,23 @@ func (s ServerForClient) GetGameState() gamestate.ClientGameState {
 func (s ServerForClient) GetGameConfig() config.ClientConfig {
 	return s.server.gameConfig.GetClientConfig()
 }
+
+func randomUniqueAssignIIGORoles(input []shared.ClientID) (shared.ClientID, shared.ClientID, shared.ClientID) {
```
lhl2617: See: https://play.golang.org/p/fBhw9ZTBK9v

internal/server/server_test.go
```diff
@@ -98,3 +98,48 @@ func TestSOMASServerFactoryInitialisesClients(t *testing.T) {
...
+				shared.ClientID(5)},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if len(tc.input) < 3 {
+				randomUniqueAssignIIGORoles(tc.input) // Only check for crash
+			} else {
+				for i := 0; i < iterations; i++ { // As its using random numbers. Run each test several times to minimise probablability
```
lhl2617: probabblability :P

internal/server/server_test.go
```diff
@@ -98,3 +98,48 @@ func TestSOMASServerFactoryInitialisesClients(t *testing.T) {
 		}
 	}
 }
+
+func TestRandomUniqueAssignIIGORoles(t *testing.T) {
```
lhl2617: as this depends on rand I'd seed the rng:
```
	rand.Seed(time.Now().UnixNano())
```

tobhil98: I wrote the tests so they'd work with the random. I read that it was difficult to tests functions with random numbers as the tests are threaded? 
Is what I've done okay, or should I try to update it?

internal/server/server_test.go
```diff
@@ -98,3 +98,48 @@ func TestSOMASServerFactoryInitialisesClients(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			if len(tc.input) < 3 {
+				randomUniqueAssignIIGORoles(tc.input) // Only check for crash
+			} else {
+				for i := 0; i < iterations; i++ { // As its using random numbers. Run each test several times to minimise probablability
+					speaker, judge, president := randomUniqueAssignIIGORoles(tc.input)
+					if speaker == judge ||
+						speaker == president ||
+						judge == president {
+						t.Errorf("Speaker: %v, Judge: %v, President: %v, should all be unique but weren't", speaker, judge, president)
```
lhl2617: I'd refactor as the golang playground link, then to test for uniqueness, write a utility function 

```
func lstHasUniqueClientIDs(lst []shared.ClientID) bool {
    // set to contain what we've seen so far
    s := map[shared.ClientID]interface{}
    for _, id := range lst {
        if _, ok := s[id]; ok { 
            return false
        }
        s[id] = nil
    }
    return true
}

File: internal/server/server_test.go (hunk 9 above)
Comment (lhl2617) [reply internal/server/server_test.go]: > it was difficult to tests functions with random numbers as the tests are threaded?

Yes it is indeed hard to test with rand stuff. Not really because of the threaded reason (because we currently don't run them in parallel)--more because we don't know what sequence the tests are run in. lots of details in here that I myself am not clear with, however.

Usually we would inject the `rand` functions we use in the function (e.g. if we use `rand.Shuffle`, we'd pass in instead as a parameter a `shuffler` function that we can mock in the tests.

This is a simple test for relatively simple functionality--seeding the RNG by time is fine--it might cause flaky tests, but eh, at this point, I'm fine with that. the main point is to prevent regression (i.e. someone accidentally touching that code and failing your requirements)

In the below I suggested a _property_-based test that instead checks for uniqueness in the lst.

File: internal/server/server.go (hunk 7 above)
Comment (tobhil98) [reply internal/server/server.go]: Oh, my question seems to have been deleted. What would you recommend that I do with the error as createSOMASServer does not use it? Just panic?
I was hoping that this problem would have been picked up before this point if there weren't any clients added

File: internal/server/server_test.go (hunk 10 above)
Comment (tobhil98) [reply internal/server/server_test.go]: Thanks. I'll add this

File: internal/server/server.go (hunk 7 above)
Comment (lhl2617) [reply internal/server/server.go]: https://play.golang.org/p/AG6u36fe6xl <- this is cleaner :) uses more memory though

Comment (tobhil98) [reply internal/server/server.go]: Hope it's better now

Comment (tobhil98) [reply internal/server/server.go]: I'll add a few more tests

File: internal/server/server.go [hunk 11]
```diff
@@ -172,3 +176,22 @@ func (s ServerForClient) GetGameState() gamestate.ClientGameState {
 func (s ServerForClient) GetGameConfig() config.ClientConfig {
 	return s.server.gameConfig.GetClientConfig()
 }
+
+func getNRandValuesFromLstUniqueWherePossible(input []shared.ClientID, n int) ([]shared.ClientID, error) {
+	if len(input) == 0 {
+		return nil, fmt.Errorf("RIP, empty list")
```
lhl2617:
```suggestion
		return nil, errors.Errorf("Empty list")
```

internal/server/server.go
```diff
@@ -172,3 +176,22 @@ func (s ServerForClient) GetGameState() gamestate.ClientGameState {
 func (s ServerForClient) GetGameConfig() config.ClientConfig {
 	return s.server.gameConfig.GetClientConfig()
 }
+
+func getNRandValuesFromLstUniqueWherePossible(input []shared.ClientID, n int) ([]shared.ClientID, error) {
```
lhl2617:
```suggestion
func getNRandClientIDsUniqueIfPossible(input []shared.ClientID, n int) ([]shared.ClientID, error) {
```

internal/server/server.go
tobhil98: Yeah, sorry. Stupid...

internal/server/server.go
```diff
@@ -70,7 +71,10 @@ func createSOMASServer(
 	}
 
 	availableRules, rulesInPlay := rules.InitialRuleRegistration(gameConfig.IIGOConfig.StartWithRulesInPlay)
-
+	initRoles, err := getNRandValuesFromLstUniqueWherePossible(clientIDs, 3)
+	if err != nil {
+		panic(err)
```
lhl2617: Instead of panicking let's change the return type of `createSOMASServer` to `(Server, error)` and return the error.
Error should be `errors.Errrof("Cannot initialise IIGO roles, no clients registered: %v", err)`

`NewSOMASServer` should also be extended with the return type of `(Server, error)`. Then, in `main.go:main()`, 
```
s, err := server.NewSOMASServer(gameConfig)
if err != nil {
    log.Fatalf("Failed to get initialise SOMASServer: %v", err)
}
```
`log.Fatalf` calls `os.Exit(1)` right after logging that stmt.


You also need to change `main_js.go:RunGame`.

```
	s := server.NewSOMASServer(gameConfig)
```
to
```
	s, err := server.NewSOMASServer(gameConfig)
	if err != nil {
		return js.ValueOf(map[string]interface{}{
			"error": convertError(errors.Errorf("Failed to get initialise SOMASServer: %v", err)),
		})
	}
```

internal/server/server_test.go
```diff
@@ -98,3 +98,91 @@ func TestSOMASServerFactoryInitialisesClients(t *testing.T) {
...
+	cases := []struct {
+		name      string
+		input     []shared.ClientID
+		unique    bool
+		retLength int
+	}{
+		{
+			name:      "RandomAssign 2 entries",
+			input:     []shared.ClientID{shared.ClientID(0), shared.ClientID(1)},
+			unique:    false,
```
lhl2617: nitpick: you dont really need `unique` here do you :) (unique is just len(input) < retLength)

internal/server/server.go
tobhil98: Hmm... So that means I'll need to return an error from NewSOMASServer, which I would then also need to capture in main_js.go?

lhl2617: sorry for asking so much for such a little change. because this is a failing point, it's really important to return **and handle** errors. Otherwise you'll just get an error like in the description of #330 

If we make it work with good practices from the start, we'll not need to spend much time debugging it.

I hope you understand :)

lhl2617: If you don't want to change the return types I can do that for you. just tell me

lhl2617: done in #374 . Do rename the error to `errors.Errrof("Cannot get random clients to initialise IIGO roles: %v", err)`

lhl2617 (approved):
aiite sorry for the trouble

internal/server/server.go
```diff
@@ -70,6 +71,10 @@ func createSOMASServer(
 	}
 
 	availableRules, rulesInPlay := rules.InitialRuleRegistration(gameConfig.IIGOConfig.StartWithRulesInPlay)
+	initRoles, err := getNRandClientIDsUniqueIfPossible(clientIDs, 3)
+	if err != nil {
+		return nil, errors.Errorf("Cannot initialise IIGO roles, no clients registered: %v", err)
```
lhl2617:
```suggestion
		return nil, errors.Errorf("Cannot initialise IIGO roles: %v", err)
```

PR #375, opened 2021-01-09
[IIGO] Made default sanction score a parameter

Neelesh99:
# Summary

Made default sanction score a parameter so we can change it with Sim. Also renamed IIGOSanctionScore to RuleSanctionPenalty, because it was too similar to IIGOSanctionsScore. (There's an extra s)

### DIFF
--- a/internal/server/iigointernal/judiciary_test.go
+++ b/internal/server/iigointernal/judiciary_test.go
@@ -565,7 +565,7 @@ func TestCreateBroadcastsForSanctionThresholds(t *testing.T) {
 						T:           shared.CommunicationInt,
 						IntegerData: 0,
 					},
-					shared.IIGOSanctionScore: {
+					shared.RuleSanctionPenalty: {
 						T:           shared.CommunicationInt,
 						IntegerData: 1,
 					},
@@ -600,7 +600,7 @@ func TestCreateBroadcastsForRuleViolationPenalties(t *testing.T) {
 						T:        shared.CommunicationString,
 						TextData: "inspect_allocation_rule",
 					},
-					shared.IIGOSanctionScore: {
+					shared.RuleSanctionPenalty: {
 						T:           shared.CommunicationInt,
 						IntegerData: int(50),
 					},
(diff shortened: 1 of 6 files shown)


internal/server/iigointernal/judiciary_test.go
```diff
@@ -1977,8 +1977,10 @@ func defaultInitJudiciary() judiciary {
 		sanctionRecord:        map[shared.ClientID]shared.IIGOSanctionsScore{},
 		sanctionThresholds:    map[shared.IIGOSanctionsTier]shared.IIGOSanctionsScore{},
 		ruleViolationSeverity: map[string]shared.IIGOSanctionsScore{},
-		gameConf:              &config.IIGOConfig{},
-		gameState:             &gamestate,
+		gameConf: &config.IIGOConfig{
+			DefaultSanctionScore: 5,
+		},
```
lhl2617, line 1982: image

Neelesh99: How do you have these memes ready to go XD

params.go
```diff
@@ -440,6 +446,7 @@ func parseConfig() (config.Config, error) {
 		InspectBallotActionCost:         shared.Resources(*iigoInspectBallotActionCost),
 		InspectAllocationActionCost:     shared.Resources(*iigoInspectAllocationActionCost),
 		AppointNextPresidentActionCost:  shared.Resources(*iigoAppointNextPresidentActionCost),
+		DefaultSanctionScore:            shared.IIGOSanctionsScore(*iigoDefaultSanctionScore),
```
lhl2617, line 449: idk why the sequence does not match that of the vars above but i'd strive to match it so if you forgot to put it in you know immediately from the seq

params.go
```diff
@@ -346,6 +346,12 @@ var (
 		"Length of the term for the Judge",
 	)
 
+	iigoDefaultSanctionScore = flag.Uint(
```
lhl2617: this implies that the sanctionscore can be changed mid game right

internal/server/iigointernal/judiciary_test.go
Neelesh99: Yeah it's just for a test so it'll always use that value to test that function.

lhl2617: at fb, we had macros to easily get memes in. I miss that so much. I just google now.

params.go
Neelesh99: Yup so judges have the option to upload their own custom scoring for particular rules that they thing should be scored more/less severely than the default. This value is what we revert to if the client judge doesn't give us a custom value for a particular rule.

lhl2617 (approved):
preapproving but i hope you can arrange the sequence of the params

params.go
Neelesh99: Yeah fair enough, updated in latest commit

PR #378, opened 2021-01-10
Client pres

P-A-Ha:
# Summary

added some logic for a very very improbable div0
## Additional Information

## Test Plan


lhl2617:
can you instead open one that pulls into #311 (JamesTev:main)?

P-A-Ha:
> can you instead open one that pulls into #311 (JamesTev:main)?#382 
oops that was by accident ;p

PR #380, opened 2021-01-10
 [INFRA/CORE] Handle invalid values in takeResources, giveResources

lhl2617:
# Summary
As title. Handle `NaN`s and negative values because _no one can be trusted_.

## Additional Information

Handled errors from `giveResources` by logging when error return is not available, returning a wrapped error otherwise.

## Test Plan

In CI we trust.

### DIFF
--- a/internal/server/iigo.go
+++ b/internal/server/iigo.go
@@ -5,6 +5,7 @@ import (
 	"github.com/SOMAS2020/SOMAS2020/internal/common/rules"
 	"github.com/SOMAS2020/SOMAS2020/internal/common/shared"
 	"github.com/SOMAS2020/SOMAS2020/internal/server/iigointernal"
+	"github.com/pkg/errors"
 )
 
 // runIIGO : IIGO decides rule changes, elections, sanctions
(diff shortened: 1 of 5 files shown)


internal/server/iigo.go
```diff
@@ -89,7 +90,10 @@ func (s *SOMASServer) runIIGOAllocations() error {
 		allocation := v.RequestAllocation()
 
 		if allocation <= s.gameState.CommonPool {
-			s.giveResources(clientID, allocation, "allocation")
+			err := s.giveResources(clientID, allocation, "allocation")
+			if err != nil {
+				return errors.Errorf("Failed to give resources: %v", err)
```
Neelesh99, line 95: Just to confirm, will the program exit if it get's to here?

lhl2617: yes. do you want it to not when it is trying to give NaN/negative?

Neelesh99: Oh right sorry, I thought it gave an error even when there isn't enough in common pool

Neelesh99: If it only produces errors on NaN/negative that's fine with me

Neelesh99 (approved):
LGTM

PR #381, opened 2021-01-10
[INFRA/CORE] Cleanup

lhl2617:
# Summary
- [team4] Change a global slice into a global slice literal (effectively making it a "constant")
- Remove RIP in error.

## Test Plan

In CI we trust.


Spruudis (approved):
RIP RIP in error

PR #382, opened 2021-01-10
[VIS] add IIGOPayments tab

ramon15749:
# Summary

Add IIGO Payments tab that shows actual/ expected tax, allocation, and sanction

ps. sorry for the code, i dont know typescript, dont kill me pls

## Additional Information

image

## Test Plan

tested the processPaymentsData function


website/src/components/Visualisations/IIGOPayments/IIGOPayments.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { ProcessedTaxData } from './Util/IIGOPaymentsTypes'
+import { processPaymentsData } from './Util/ProcessedPaymentsData'
+import { OutputJSONType, TeamName } from '../../../consts/types'
+
+const Payments = (props: { output: OutputJSONType }) => {
+  const [data, setData] = useState(processPaymentsData(props.output))
+
+  useEffect(() => {
+    setData(processPaymentsData(props.output))
+  }, [props.output])
```
lhl2617: dont need state :) only set once. just do
```
const { output } = props
```
and you're good to go

website/src/components/Visualisations/IIGOPayments/IIGOPayments.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const teams = ['Team1', 'Team2', 'Team3', 'Team4', 'Team5', 'Team6']
```
lhl2617: this is available in `output.AuxInfo`

website/src/components/Visualisations/IIGOPayments/Util/ProcessPaymentsData.test.ts
```diff
@@ -0,0 +1,2163 @@
+import { processPaymentsData } from './ProcessedPaymentsData'
+
+const testInput = {
```
lhl2617, line 3: @Darrekt do we already have some sample test inputs?

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
```diff
@@ -0,0 +1,116 @@
...
+            actualTax: value.actualTax,
+            expectedTax: value.expectedTax,
+            actualAlloc: value.actualAlloc,
+            expectedAlloc: value.expectedAlloc,
+            actualSanction: value.actualSanction,
+            expectedSanction: value.expectedSanction,
+        })
+    })
+
+    return retData
```
lhl2617, line 78:
```
return TaxInfo.map((value, key) => {
    ....
}
```

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
```diff
@@ -0,0 +1,116 @@
...
+                switch (type) {
+                    case 'IslandAllocation':
+                        {
+                            const team =
+                                TeamName[
+                                    teamAction.ClientID as keyof typeof TeamName
+                                ]
+                            const current: TaxData | undefined = TaxInfo.get(
+                                team
+                            )
```
lhl2617: you can move this above L35 so that we don't need to do it for each case :)

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
```diff
@@ -0,0 +1,116 @@
...
+        if (exchanges) {
+            exchanges.forEach((teamAction) => {
+                const type = teamAction.Pairs[0].VariableName
+                switch (type) {
+                    case 'IslandAllocation':
+                        {
+                            const team =
+                                TeamName[
+                                    teamAction.ClientID as keyof typeof TeamName
+                                ]
```
lhl2617: hmm, don't think this is the best way. I think you can cast directly?

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
```diff
@@ -0,0 +1,116 @@
...
+                            const current: TaxData | undefined = TaxInfo.get(
+                                team
+                            )
+                            if (teamAction.Pairs[0].Values[0] !== 0) {
+                                console.log(turnNumber, teamAction)
```
lhl2617: why?

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
```diff
@@ -0,0 +1,116 @@
...
+                            }
+                        }
+                        break
+                    default:
+                        break
+                }
+            })
+        }
+    })
+    console.log(TaxInfo)
```
lhl2617: can use `console.debug` :) (you'll have to turn on all log levels on your console however)

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    const TaxInfo: Map<TeamName, TaxData> = new Map<TeamName, TaxData>([
+        [1, empty()],
+        [2, empty()],
+        [3, empty()],
+        [4, empty()],
+        [5, empty()],
+        [6, empty()],
+    ])
```
lhl2617, line 30: does anyone else know a cleaner way of doing this? :O 

I would do it in a different way in which I `map` then `reduce` (for the whole function), but hey, if this works, it works :D

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    IIGOHistory.forEach(([turnNumber, exchanges]) => {
+        if (exchanges) {
```
lhl2617, line 33: there would be an undefined exchange? :O

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
ramon15749: I dont know if i did something wrong but when i tried `TaxInfo.map` i got the error saying map function doesnt exist on type 'Map' ..?

website/src/components/Visualisations/IIGOPayments/Util/ProcessPaymentsData.test.ts
ramon15749: @Darrekt used some thing similar in his testing, but there is a lot of wrong/weird values inside so I change a few things in mine :(

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
ramon15749: to be honest, I am not sure either but it is done this way [here](https://github.com/SOMAS2020/SOMAS2020/blob/main/website/src/components/Visualisations/Transactions/Util/ProcessTransactionData.ts#L15)  so I just follow

lhl2617: ah ok :O

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
lhl2617: ah, nevermind. it's a map. you'd need to use the spread operator or something to do it. anyway--it works, and I don't want to block this.

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    IIGOHistory.forEach(([_, exchanges]) => {
+        if (exchanges) {
+            exchanges.forEach((teamAction) => {
+                const team = TeamName[teamAction.ClientID]
+                const current: TaxData | undefined = TaxInfo.get(team)
+                const type = teamAction.Pairs[0].VariableName
+                switch (type) {
+                    case 'IslandAllocation':
+                        if (current) {
```
lhl2617: seems like every branch of the cases do `if (current)`, why not do it above L38 once :)

lhl2617 (approved):
I'm OK with this but would like if @Darrekt and @yp717 could have a deeper read--part of the code seems to have come from @Darrekt.
Preapproving

website/src/components/Visualisations/IIGOPayments/Util/ProcessPaymentsData.test.ts
lhl2617: ah, that's fine then, it's just a bit yuuge

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
yp717: I would have done it with map and reduce too can't think of a nicer way :O if it works it works :D

ramon15749: so should i change it to map and reduce, or should i just keep it this way since its (kind of) working?

MikeIsMike:
just wondering what "expectedAlloc" means? Is it the allocation that an agent requested? Or is it the allocation that the president granted to an agent. And what about "actualAlloc"? Could you change them to more informative names?

ramon15749:
> just wondering what "expectedAlloc" means? Is it the allocation that an agent requested? Or is it the allocation that the president granted to an agent. And what about "actualAlloc"? Could you change them to more informative names?

@MikeIsMike  I just changed the legend naming, let me know if this makes more sense. I dont english crei ;(

website/src/components/Visualisations/IIGOPayments/Util/ProcessedPaymentsData.ts
yp717: keep it its fine we have basically no test coverage on the front end anyway it's fine for this sort of scale project I think :)

yp717 (approved):
lgtm and love that you're using react hooks :)

PR #369, opened 2021-01-09
[Agent] Team 4 Outgoing custom client

MikeIsMike:
# Summary

Please provide a summary of the change.

## Additional Information

Auxiliary information goes here.

## Test Plan

- Add information on how you tested your changes.

PR #384, opened 2021-01-10
[CLIENT/Team4] 1 out of 4

akowal3:
# Summary

Team 4 client implementation

## Test Plan

- Added tests where time and manpower allowed.


internal/clients/team4/client.go
```diff
@@ -16,51 +18,102 @@ func init() {
 	baseclient.RegisterClientFactory(id, func() baseclient.Client { return NewClient(id) })
 }
 
-// NewClient is a function that creates a new empty client
-func NewClient(clientID shared.ClientID) baseclient.Client {
-	iigoObs := &iigoObservation{
+func newClientInternal(clientID shared.ClientID, testing *testing.T) client {
```
lhl2617, line 21: that's a very strange way of testing :O, is it not possible to unit test the judge separately? Including the `testing` package in production is something that rarely happens.

internal/clients/team4/trust.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (t *trust) expectedTrustSum() float64 {
+	return 0.5 * float64(len(t.trustMap))
+}
+
+//normalise ensures that trust values are always in range [0,1]
+func (t *trust) normalise() {
+	// diff is percentage to increase trust for clientID ie. diff in range [0,1]
+	if len(t.trustMap) > 0 {
+		normaliseCoef := t.expectedTrustSum() / t.totalTrustSum()
```
NGovani, line 60: divide by 0 check, if you trust no one this would be 0

internal/clients/team4/iifo.go
```diff
@@ -0,0 +1,140 @@
...
+		totalCoordinateY += prediction.PredictionMade.Confidence * prediction.PredictionMade.CoordinateY
+		totalMagnitude += prediction.PredictionMade.Confidence * prediction.PredictionMade.Magnitude
+		totalTimeLeft += uint(math.Round(prediction.PredictionMade.Confidence)) * prediction.PredictionMade.TimeLeft
+		totalConfidence += prediction.PredictionMade.Confidence
+	}
+
+	// Finally get the final prediction generated by considering predictions from all islands that we have available
+	// This result is currently unused but would be used in decision making in full implementation
+	finalPrediction := shared.DisasterPrediction{
+		CoordinateX: totalCoordinateX / totalConfidence,
```
NGovani, line 123: div by 0

internal/clients/team4/client.go
```diff
@@ -153,10 +219,10 @@ func (c *client) decideRuleDistance(ruleMatrix rules.RuleMatrix) float64 {
 			} else if currentAuxValue == 1 {
 				// TODO: ACTUALLY IMPLEMENT THESE CONDITIONS
 				// >0 condition
-				distance += 10000
+				distance += math.Abs(idealValue-actualValue) / idealValue
```
NGovani: div by 0

NGovani (changes requested):

internal/clients/team4/judge.go
```diff
@@ -102,29 +125,45 @@ func (j *judge) HistoricalRetributionEnabled() bool {
...
 
 // DecideNextPresident returns the ID of chosen next President
 // OPTIONAL: override to manipulate the result of the election
-func (j *judge) DecideNextPresident(winner shared.ClientID) shared.ClientID {
-	// overloaded
-	j.logf("hello world %v", winner)
-	return id
-}
 
 func (j *judge) logf(format string, a ...interface{}) {
```
lhl2617, line 168: ah--you're using `t` here to logf. I prefer you instead pass in a logger function. `t.Log` is a function pointer you can pass in the judge. Your judge can then take in any logger function it wants.

lhl2617: This was in #369--I missed this part. Well, it works now so I won't block you, but I'll create an issue

internal/clients/team4/trust.go
akowal3: yeah, although we maintain the average trust at 0.5, so this situation will not occur. I'll add a check for the peace of mind

internal/clients/team4/judge.go
akowal3: if I recall correctly I had some problems with logger in testing, hence this approach.

internal/clients/team4/client.go
lhl2617: ignore for now. would love to see a PR resolving this. tracking in #386

internal/clients/team4/judge.go
lhl2617: this is a very side-effecty way of testing unfortunately, but I don't want to block this becuase of that (since the change bringing this in was in #369). Tracking in #386

internal/clients/team4/client.go
MikeIsMike: sorry PR will be updated soon

NGovani (approved):
LGTM

internal/clients/team4/iifo.go
MikeIsMike: resolved

PR #385, opened 2021-01-10
[CLIENT/Team4] 2 out of 4

akowal3:
# Summary

Do not merge before #384 

- Added trust datastructure and helper functions
- gamestate is now passed to judge/speaker (major bugfix)

PR #397, opened 2021-01-11
[CLIENT/Team4] removed testing object from prod code

### LINKED ISSUE #386: [TEAM4] Client using `testing` package in non-test client code
`judge` is using a testing object to log to it. This is not a great pattern -- you should instead pass in a logger function.

akowal3:
Fixed #386 
Fixed the test LH mentioned in #385


lhl2617 (approved):
Thanks!

PR #399, opened 2021-01-11
[INFRA/IIGO] Amended Default Monitoring Result Default

preetl:
# Summary

Corrected the logic in IIGO role monitoring. If a role is not monitored the default result should be 'true' i.e. no wrong doing was found. I've looked through client code and it seems that this is how everyone has assumed the logic is already so this doesn't mess up any client implementations.

## Additional Information

N/A

## Test Plan

Still passing!


NGovani (approved):

Spruudis (approved):
LGMT good catch!

PR #419, opened 2021-01-12
[IIGO/Vis] 1AM patches tend to leave scratches

Spruudis:
Missed case last night of IIGO quitting early, this solves that


preetl (approved):
thank uuuuu

PR #422, opened 2021-01-12
[TEAM 3] Couldn't repro panic, but patched so that it doesn't get tot that point

Neelesh99:
# Summary

Sorry @E-Mac27 I couldn't reproduce the panic you were getting, but looking at the call stack I've patched it so that it won't get into that situation.

### DIFF
--- a/internal/clients/team3/dynamics/ruleanalysis.go
+++ b/internal/clients/team3/dynamics/ruleanalysis.go
@@ -88,7 +88,7 @@ func FindClosestApproach(ruleMatrix rules.RuleMatrix, namedInputs map[rules.Vari
 			return laceOutputs(bestPosition, ruleMatrix, droppedInputs, namedInputs)
 		}
 	}
-	return map[rules.VariableFieldName]Input{}
+	return namedInputs
 }
 
 func Shift(ruleMatrix rules.RuleMatrix, namedInputs map[rules.VariableFieldName]Input) (newMatrix rules.RuleMatrix, edited bool) {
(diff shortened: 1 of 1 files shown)


internal/clients/team3/dynamics/ruleanalysis.go
```diff
@@ -469,6 +469,10 @@ func decodeWithConst(rm rules.RuleMatrix, values map[rules.VariableFieldName][]f
 func evaluateSingle(rm rules.RuleMatrix, values map[rules.VariableFieldName][]float64) (outputVec mat.VecDense, success bool) {
 	varVect := decodeWithConst(rm, values)
 	if varVect != nil {
+		nrows, ncols := rm.ApplicableMatrix.Dims()
+		if nrows == 0 || ncols == 0 {
+			return mat.VecDense{}, false
+		}
```
E-Mac27, line 475: I'm assuming this has to be a square matrix? Cause if it isn't - I'm scared that it might affect your foraging if a non-square matrix exists.

Edit: not foraging. Rule analysis xD

E-Mac27 (approved):
Sorry! Sad you couldn't replicate it. I'm hoping this doesn't affect foraging for you guys in general.

internal/clients/team3/dynamics/ruleanalysis.go
Neelesh99: I don't believe there is a dependency on square matrices, or at least I hope not.

PR #426, opened 2021-01-12
IIGO Negative Allocation. Quick fix

E-Mac27:
# Summary
As title.

# Tests
All test pass


Neelesh99:
Looks good, should we also put checks on tax and sanction payments. Just incase?

lhl2617 (approved):
image


certified negative

E-Mac27:
> Looks good, should we also put checks on tax and sanction payments. Just incase?

I've ran the sim for 1500 turns and this didn't cause an issue but I dont mind being more cautious and adding it.

Edit: Only allocation throws an error for some reason. Sanction and Tax just logs and assigns 0 value. So checks are already in place.

internal/server/iigo.go
```diff
@@ -90,7 +90,10 @@ func (s *SOMASServer) runIIGOAllocations() error {
 	allocationMap := make(map[shared.ClientID]shared.Resources)
 	for clientID, v := range clientMap {
 		allocation := v.RequestAllocation()
-
+		if allocation < 0 {
+			s.logf("Negative allocation of %v by %v. Changing allocation to 0", allocation, clientID)
```
lhl2617: let's save some time

```suggestion
		if allocation < 0 || math.IsNaN(allocation) {
			s.logf("Invalid allocation of %v by %v. Changing allocation to 0", allocation, clientID)
```

PR #430, opened 2021-01-13
[Sim/Vis/GameState/SORRY] Vis feature metrics

Eirikalb:
# Summary

Provides some new metrics, added a div zero check to computeUnmitigatedDisasterEffects()
New Metrics:
- Rules broken
- Commonpool mitigation of disasters
- Number of disasters survived
- Impact of disasters survived
Also changed some colours for the foraging graph to make a clear distinction between fishing and deerhunting
image
image

## Additional Information

## Test Plan


internal/common/disasters/environmentdefinition.go
```diff
@@ -32,6 +32,7 @@ type disasterParameters struct {
 type DisasterReport struct {
 	Magnitude shared.Magnitude
 	X, Y      shared.Coordinate
+	Effects   DisasterEffects
 }
```
lhl2617, line 36: Wondering if this is copied correctly as part of GameState?

lhl2617: structs are reference types

website/src/components/Visualisations/Metrics/MetricEntries.ts
```diff
@@ -122,6 +129,70 @@ const returnsFromCriticalMetricCollection = (data: OutputJSONType) => {
     return metrics
 }
 
+const rulesBrokenPerTeamCollection = (data: OutputJSONType) => {
+    const metrics = emptyMetrics()
+    // Since IIGOHistories is repeated, take the one from the LAST GameState and
+    // do Object.entries to make it iterable. List of array'ed tuples.
+    data.GameStates.forEach((gameState) => {
+        if (gameState.RulesBrokenByIslands !== null) {
+            Object.entries(gameState.RulesBrokenByIslands).forEach((island) => {
```
lhl2617, line 138: Prefer Object.entries(obj).forEach([k, v]) => ...
with well named k and v

lhl2617 (approved):
ok if checked copy works correctly

internal/common/disasters/environmentdefinition.go
Eirikalb: It is in the gamestate but it seems that the first entry is null for the Effects fields...

Eirikalb: The copy works correctly, but I the DisasterReport does not have a default value so the first round copies nulls.

PR #431, opened 2021-01-13
late night work :|

E-Mac27:
# Summary
Change params to make more sense.

Islands survive for 100+ rounds. But small chance of islands also wiping :)
image

PR #433, opened 2021-01-13
Revert "late night work :|"

lhl2617:
Reverts SOMAS2020/SOMAS2020#431

This is causing OOM errors.

Let's revert for now and I will figure out a solution after my exam. For now, just copy params.go from here.


Neelesh99 (approved):
Not sure what OOM is, but you've my blessing

PR #437, opened 2021-01-13
Making sims run for 50 turns instead of 20

E-Mac27:
# Summary
As title. I ran it like 10 times and it didn't throw OOM. 

## Additional Information
image

When running `yarn start`:
image

## Test Plan
Tests all pass
image


lhl2617 (approved):
YAS

PR #440, opened 2021-01-13
[Agent/Team6]Fix voting error in agent

Ning-72:


internal/clients/team6/voting.go
```diff
@@ -119,8 +119,8 @@ func (c *client) VoteForElection(roleToElect shared.Role, candidateList []shared
 	if len(candidateList) > 3 && doWeHaveRoles == false {
 		for i, role := range rolePreference {
 			if role == roleToElect {
-				insertID := append(candidateList[:i], idToSlice...)
-				candidateList = append(insertID, candidateList[i:]...)
+				insertID := append(idToSlice, candidateList[i:]...)
```
Neelesh99: As @lhl2617 mentioned in an earlier PR, this is slightly dangerous since if idToSlice has enough space this append will mutate idToSlice. Therefore, when the for loop comes back around again you will be operating on a different idToSlice. Please deepCopy or change the logic so that you only ever assign append(x, y) back to x to guarantee correct operation.

Neelesh99 (changes requested):
A small issue with the append

internal/clients/team6/voting.go
Ning-72: Oh I see, would it be ok to break the loop after the append operation? then idToSlice would be used once I think.

Neelesh99: Personally I have a gripe against breaking loops, but that's just me so it's fine to do that. But make sure wherever you do it that when you do the append you assign the result to the slice you're appending to. e.g. idToSlice = append(idToSlice, y...)

Ning-72: I put the operation outside the loop, there should be no conflicts I think, thanks!

Neelesh99: Not sure if this is stable, but it's ok I guess

Neelesh99 (approved):
Looks good for now, not sure about that append still

Issue #439, opened 2021-01-13
Some agents has put same people in the ballot multiple times


ramon15749:
Some agents (Team3, Team4 and Team6 (might be others but this is what I found)) has put the same people in the ballot multiple times.

![image](https://user-images.githubusercontent.com/35395504/104504470-66bfa800-55da-11eb-91b6-32d27cdcdf4a.png)

[error.zip](https://github.com/SOMAS2020/SOMAS2020/files/5810835/error.zip)

ramon15749:
[output.zip](https://github.com/SOMAS2020/SOMAS2020/files/5810513/output.zip)

preetl:
Team 3 has fixed this issue (for our team)! With thanks to @Neelesh99 and @ramon15749

Ning-72:
Sorry, I think it's fixed now #440

PR #441, opened 2021-01-13
[VOTING/TEAM 3] Multiple entries of teams in ballot card. 

Neelesh99:
# Summary

There were 2 separate bugs which cause clients to appear multiple times in ballots, causing skewing of results.
Bug 1: The candidate list was passed by reference which meant that if a team accidentally mutated it for one elected, a mutated list (often with duplicitous entries) would be passed to the next election.
Bug 2: Team 3 (among others #439 ) have a bug where they put multiple entries of the same agent in their ballots. These entries are double counted and skew results.

Team 3 has fixed our code so it doesn't put multiple entries of the same client id in the ballot paper.


preetl (approved):
fabulous!

PR #444, opened 2021-01-13
[TEAM4/CLIENT] Voting fix

tobhil98:
# Summary
- Fix voting issue reported in #439


internal/clients/team4/client.go
```diff
@@ -304,9 +304,15 @@ func (c *client) VoteForElection(roleToElect shared.Role, candidateList []shared
 	returnList := []shared.ClientID{}
 	for i := 0; i < len(candidateList); i++ {
 		trustScore := c.trustMatrix.GetClientTrust(candidateList[i]) //c.internalParam.agentsTrust[candidateList[i]]
+		_, ok := trustToID[trustScore]
+		for ok {
+			trustScore += 0.000001 // Tiny increment to make it unique
```
Neelesh99, line 309: A very interesting way to do this XD

Neelesh99 (approved):
Pre-approving since the changes themselves seem fine. Have you guys run and checked it doesn't create duplicates?

internal/clients/team4/client.go
tobhil98: Yeah... I didn't want to rank them equally anyway, so just made it unique. Our problem arose when all clients have the same trust at the start. Also, I thought it would be safer than rewriting it and perhaps introducing bugs. This should give roughly the same simulation results as before.

tobhil98:
> Pre-approving since the changes themselves seem fine. Have you guys run and checked it doesn't create duplicates?

Yeah, I've run it in debug mode and it seemed fine

